PR #5, opened 2022-02-18
feat: enable PDF output

fghaas:
Add mkdocs-pdf-export-plugin, which (if Mkdocs is running with the `DOCS_ENABLE_PDF_EXPORT` environment variable set to `1`) renders PDF versions of all pages, and adds a download button right next to the edit button on every page.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -11,10 +11,15 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install packages
+        run: |
+          sudo apt-get install weasyprint
+      - name: Install Python dependencies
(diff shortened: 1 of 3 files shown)


evrardjp (changes requested):
Shouldn't we exercise this feature in CI directly, and add the env var in the github action?

evrardjp:
I see.

IMO if you enable the feature, you want to test it. Either through manual review or automated review.
I suppose the feature is important to you, so I assume you run this on your machine (if not in CI).
Do you expect others to also review this? If true, please mention it in the contributor guide.

evrardjp (approved):
The pdfs look okay to me.

There is this one page empty at the start of the contributing guide. Not sure if it's intended or not, but it's good enough for me. We can improve the templating later. The framework is in place, which is all what matters as of now.

Issue #6, opened 2022-02-18
Uploaded build is named build-version not build-version-SHA


evrardjp:
More a nuisance than anything, but fetching the artifacts results in file always having the same name. Having multiple filenames (one per sha) or (alternatively) having a way to _view_ the artifacts direct from gh would help.

fghaas:
It's not build-version; it's build-pythonversion. The only point is to be able to distinguish between build environments.

PR #7, opened 2022-02-18
Add merge SHA into build artifact

evrardjp:
When building artifact, by default the name is always build-version.
This is a problem if you're working on multiple PRs for which
you download the artifact: It's now harder to know who is who.

This fixes it by adding the SHA of the virtual merge commit into
the filename of the artifact.

Closes: #6

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Upload build
         uses: actions/upload-artifact@v2
         with:
-          name: build-${{ matrix.python-version }}
+          name: build-${{ matrix.python-version }}-${{ github.sha }}
           path: site/
     strategy:
       matrix:


.github/workflows/tox.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Upload build
         uses: actions/upload-artifact@v2
         with:
-          name: build-${{ matrix.python-version }}
+          name: build-${{ matrix.python-version }}-${{ github.sha }}
```
fghaas, line 26: Given the fact that every change is on a topic branch (since we've protected `main`), and also few people remember the SHAs of head commits of topic branches, maybe include `{{ github.ref }}` instead, or both (as in `build-${{ matrix.python-version }}-${{ github.ref }}-${{ github.sha }}`)

fghaas:
Also, please run `git config core.hooksPath .githooks` in your local checkout and add a `build:` prefix to your commit message.

.github/workflows/tox.yml
evrardjp: ref is containing forward slashes, updating to ref_name, which is more convenient.

evrardjp: oh well, ref_name also contains it. Gonna remove ref. It's fine enough.

fghaas:
Sorry, accidentally pushed to your fork rather than mine. Will force-push to undo my change; hope that's OK with you.

evrardjp:
> Sorry, accidentally pushed to your fork rather than mine. Will force-push to undo my change; hope that's OK with you.

No problem at all :)

evrardjp:
If you're happy with the result, I am happy. It's not perfect, but ti's an improvement over what's existing.

.github/workflows/tox.yml
fghaas: Not quite sure what you mean; I just tried this and `github.ref_name` does appear to contain just the branch name:

```
name: "build-${{ matrix.python-version }}-${{ github.ref_name }}-${{ github.sha }}"
```

evrardjp: it does contain `something/branchname`. So that kinda is annoying, as `/` is recognized as directory separator. I don't want to deal with substitutions.

fghaas: Er, no it doesn't? See https://github.com/fghaas/docs/actions/runs/1875290546

evrardjp: https://github.com/citynetwork/docs/actions/runs/1865627547

evrardjp: Correct.

fghaas:
Hnngh. Neglected to look into the commit message one more time; now we've got a commit message that contradicts the change (as it purports to include the branch name when it doesn't). Ah well, not gonna start force-pushing to fix that. :)

evrardjp:
I am sorry. We'll make do indeed.

PR #9, opened 2022-03-08
feat: Add how-to guides for Barbican

fghaas:
Expand the how-to guides section with information on

* How to use Barbican for generic secret storage
* How to share and unshare Barbican secrets
* How to use Barbican for encrypted volumes
* How to use Barbican for HTTPS termination in Octavia
* How to use Barbican for SSE-C against Ceph-backed S3

### DIFF
--- a/docs/How-To_Guides/Object_storage/s3-credentials.md
+++ b/docs/How-To_Guides/Object_storage/s3-credentials.md
@@ -1,0 +1,53 @@
+# Working with S3-compatible credentials
+
+When you want to interact with object storage in {{extra.brand}} using
+tools that support an Amazon S3 compatible API (such as `s3cmd`,
+`rclone`, the `aws` CLI, or the Python `boto3` library), you need an
+S3-compatible access key ID and secret key.
+
+
+## Creating credentials
+
+You can create a set of S3-compatible credentials with the following
+command:
+
+```bash
+openstack ec2 credentials create
+```
+
+This will return an `Access` and `Secret` key that you can use to
+populate the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`
+environment variables (or whichever configuration options your
+application requires).
+
+!!! note
+    Your S3-compatible credentials are always scoped to your
+    {{extra.brand}} *region* and *project*. You cannot reuse an access
+    and secret key across multiple regions or projects.
+
+    Also, your credentials are only “S3-compatible” in the sense that
+    they use the same *format* as AWS S3 does. They are never valid against
+    AWS S3 itself.
+
+
+## Listing credentials
+
+You can list any previously-created credentials with:
+
+```bash
+openstack ec2 credentials list
+```
+
+## Deleting credentials
+
+If at any time you need to delete a set of AWS-compatible credentials,
+you can do so with the following command:
+
+```bash
+openstack ec2 credentials delete <access-key-id>
+```
+
+!!! warning
+
+    Deleting a set of S3-compatible credentials will *immediately*
+    revoke access for any applications that were using it.
--- a/docs/How-To_Guides/Object_storage/sse-c.md
+++ b/docs/How-To_Guides/Object_storage/sse-c.md
@@ -1,0 +1,91 @@
+# Client-side encryption with S3 API (SSE-C)
+
+You can use objects encryption via S3 API, according to the[Amazon
+SSE-C](https://docs.aws.amazon.com/AmazonS3/latest/dev/ServerSideEncryptionCustomerKeys.html)
+specification. This means that you need to provide an encryption/decryption key
+with each request to the object.
+
+You can store the encryption key in Barbican, and provide it to the S3
+client at runtime. Below we will provide more detailed explanation
+regarding how to use this in your workloads.
+
+
+## Requirements
+
+This guide assumes familiarity with the following tools:
+
+* `python-openstackclient` (with the `python-barbicanclient` plugin),
+* `pwgen`,
+* `rclone` version 1.54 or later.
+
+
+## Create encryption details
+
+According to SSE-C specification, in order to use server-side
+encryption, any S3 client needs to provide three pieces of
+information, which it includes in the request headers for each S3
+request being made:
+
+* Encryption algorithm: the only valid option here is AES256.
+* Encryption key: a generated random key that we will store in
+  Barbican. It should be valid AES key, which means that key length
+  must be 32 bytes.
+* Encryption key checksum: MD5 checksum of the encryption key. It's
+  used for integrity checks.
+
+In order to generate encryption key and store it in Barbican, proceed
+as follows.
+
+1.  Generate secret:
+
+        secret_raw=$(pwgen 32 1)
+
+2.  Store secret in Barbican:
+
+        barbican_secret_url=$(openstack secret store --name objectSecret --algorithm aes --bit-length 256 --payload ${secret_raw} -f value -c 'Secret href')
+
+3.  Retrieve secret from Barbican:
+
+        secret=$(openstack secret get ${barbican_secret_url} -p -c Payload -f value)
+
+
+## Managing encrypted objects in S3 (with `rclone`)
+
+SSE-C encryption has been implemented/fixed with version 1.54. Prior
+rclone versions won't work here.
+
+1. Download and install the latest rclone for your distribution:
+   <https://rclone.org/downloads/>
+2. Create or retrieve your [access key ID and secret
+   key](s3-credentials.md).
+3. Create a configuration file named `~/.rclone.conf`, with the
+   following content:
+
+        [{{extra.brand|lower|replace(' ','')}}]
+        type = s3
+        provider = Ceph
+        env_auth = false
+        access_key_id = <access key id>
+        secret_access_key = <secret key>
+        endpoint = <region>.{{extra.brand_domain}}:8080
+        acl = private
+        sse_customer_algorithm = AES256
+
+4. Create an S3 bucket:
+
+        rclone mkdir {{extra.brand|lower|replace(' ','')}}:encrypted
+
+5. Sync a directory to the S3 bucket, encrypting the files it
+   contains on upload:
+
+        rclone sync ~/media/ {{extra.brand|lower|replace(' ','')}}:encrypted \
+          --s3-sse-customer-key=${secret}
+
+6. Retrieve a file from S3 and decrypt it:
+
+        rclone copy {{extra.brand|lower|replace(' ','')}}:encrypted/file.png \
+          --s3-sse-customer-key=${secret}
+
+
+For more examples on how to use rclone, please use its reference
+documentation: <https://rclone.org/docs/#subcommands>
--- a/docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
+++ b/docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
@@ -1,0 +1,91 @@
+# Generic secret storage
+
+The simplest way to use Barbican is to create and retrieve a securely
+stored, generic secret.
+
+## How to store a generic secret
+
+It is possible to store any secret data with Barbican. The command
+below will create a secret of the type `passphrase`, named `mysecret`,
+which contains the passphrase `my very secret passphrase`.
+
+```bash
+openstack secret store \
+  --secret-type passphrase \
+  -p "my very secret passphrase" \
+  -n mysecret
+```
+
+!!! note
+
+    The example output below uses {{extra.brand}}’s `Fra1` region. In other
+	regions, the secret 
+	[URIs](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier)
+	will differ.
+
+
+```
++---------------+--------------------------------------------------------------------------------+
+| Field         | Value                                                                          |
++---------------+--------------------------------------------------------------------------------+
+| Secret href   | https://fra1.{{extra.brand_domain}}:9311/v1/secrets/33ef0985-f89e-4bf0-b318-887ecac0cba |
+| Name          | mysecret                                                                       |
+| Created       | None                                                                           |
+| Status        | None                                                                           |
+| Content types | None                                                                           |
+| Algorithm     | aes                                                                            |
+| Bit length    | 256                                                                            |
+| Secret type   | passphrase                                                                     |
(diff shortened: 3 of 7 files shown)


fghaas:
@evrardjp, can you please let me know when you'll get around to reviewing this?

evrardjp:
Thanks for the ping. 
I will make sure Dmitriy also reviews it.

docs/How-To_Guides/Object_storage/s3-credentials.md
```diff
@@ -0,0 +1,53 @@
...
+	
+	Also, your credentials are only “S3-compatible” in the sense that
+	they use the same *format* as AWS S3 does. They are never valid against
+	AWS S3 itself.
+
+
+## Listing credentials
+
+You can list any previously-created credentials with:
+
```
noonedeadpunk, line 36: should we mention somewhere as Note, that Secret is shown only once during credentials creation?

docs/How-To_Guides/Object_storage/sse-c.md
```diff
@@ -0,0 +1,90 @@
...
+        env_auth = false
+        access_key_id = <access key>
+        secret_access_key = <secret key>
+        endpoint = <region>.citycloud.com:8080
+        acl = private
+        sse_customer_algorithm = AES256
+
+4. Create an S3 bucket:
+
+        rclone mkdir citycloud:encrypted
```
noonedeadpunk, line 75: rclone mkdir {{extra.brand|lower|replace(' ','')}}:encrypted

(and same below for all commands)

docs/How-To_Guides/OpenStack/Barbican/share-secret.md
```diff
@@ -0,0 +1,41 @@
+# Sharing secrets via ACLs
+
+Normally, a Barbican secret is only available to the OpenStack API
+user that created it. However, under some circumstances it may be
+desirable to make a secret available to another user.
+
+To do so, you will need
+
```
noonedeadpunk, line 8: We should include here list of UUIDs for which users need to create ACL for. They're located on a subpage in wiki

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
```diff
@@ -0,0 +1,157 @@
...
+2. You can ask the user that created the volume to [add you to the
+   Access Control List (ACL) for the
+   secret](../Barbican/share-secret.md). This will enable you to read
+   the secret, and to delete the volume using your own credentials.
+
+
+## Block device encryption caveats
+
+Once a volume is configured to use encryption and is also attached to
+an instance in {{extra.brand}}, some caveats apply that you might want
```
noonedeadpunk, line 129: Is {{extra.brand}} really applicable here?

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
```diff
@@ -0,0 +1,157 @@
...
+This means that you have two options:
+
+1. If you *do* trust us to include your instances in live migrations
+   and evacuations, even if they attach encrypted volumes, then you
+   can [add our administrative account to the Access Control List
+   (ACL) for your secrets](../Barbican/share-secret.md).
+2. If you *don’t* want to share your secrets but you still want to use
+   encrypted volumes, you should build your own mechanism or process
+   (preferably automated) so that your instances recover in case they
+   become non-functional.
```
noonedeadpunk, line 157: We should say also that anti-affinity won't be respected here. But we will provide timeframe for manual evacuation. Also I guess this could refference to "Maintenance Window Instructions for instances with encrypted volumes attached" page.

noonedeadpunk (changes requested):
Looks good in general, jsut some small nits

docs/How-To_Guides/Object_storage/s3-credentials.md
```diff
@@ -0,0 +1,53 @@
...
+openstack ec2 credentials list
+```
+
+## Deleting credentials
+
+If at any time you need to delete a set of AWS-compatible credentials,
+you can do so with the following command:
+
+```bash
+openstack ec2 credentials delete <access-key>
```
evrardjp, line 47: is that `<access-key>` the same as $AWS_ACCESS_KEY_ID?

docs/How-To_Guides/Object_storage/sse-c.md
evrardjp: Not sure if it works, but it could make sense that we have something like extra.brandsomethingsomething that contain all we need?

docs/How-To_Guides/OpenStack/Barbican/index.md
```diff
@@ -0,0 +1,25 @@
...
+* [Generic secret storage](generic-secret.md),
+* [encryption for persistent volumes](../Cinder/encrypted-volumes.md),
+* [certificate storage for HTTPS load balancers](../Octavia/tls-lb.md).
+
+To manage secrets with Barbican, you will need the `openstack` command
+line interface (CLI), and its Barbican plugin. You can install them
+both with the following commands:
+
+```bash
+pip install python-openstackclient python-barbicanclient
```
evrardjp, line 16: This is a bad idea to advise people to pip install directly. Let's just say `pip install --user` please.

docs/How-To_Guides/OpenStack/Barbican/share-secret.md
evrardjp: @noonedeadpunk  would it make sense to have _another_ page to say sharing secrets with administrators for maintenance purposes, and include the UUID of the users there?

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
evrardjp: why not? It looks okay to me.

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
```diff
@@ -0,0 +1,157 @@
...
+Live migration or evacuation *including encrypted volumes* does,
+however, require that whoever does the migration also has at least
+read access to the volume’s encryption secret.
+
+This means that you have two options:
+
+1. If you *do* trust us to include your instances in live migrations
+   and evacuations, even if they attach encrypted volumes, then you
+   can [add our administrative account to the Access Control List
+   (ACL) for your secrets](../Barbican/share-secret.md).
```
evrardjp, line 153: if we create a separate page with the procedure to grant access to our admins, maybe we can add a link here. WDYT?

evrardjp (commented):
A very good article.

A few questions before I vote.

docs/How-To_Guides/Object_storage/s3-credentials.md
fghaas: No, because that's not true. :) `openstack ec2 credentials list` always lists the existing access and secret keys defined in a tenant.

docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
```diff
@@ -0,0 +1,81 @@
...
+openstack secret store \
+  --secret-type passphrase \
+  -p "my very secret passphrase" \
+  -n mysecret
+```
+```
++---------------+--------------------------------------------------------------------------------+
+| Field         | Value                                                                          |
++---------------+--------------------------------------------------------------------------------+
+| Secret href   | https://fra1.{{extra.brand_domain}}:9311/v1/secrets/33ef0985-f89e-4bf0-b318-887ecac0cba |
```
evrardjp, line 22: In other places, you write `<region>`, would it make sense that we say, before the call, that we are using fra1 region for the example?

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
fghaas: Yes, it's correct here.

docs/How-To_Guides/Object_storage/s3-credentials.md
fghaas: Yes it is.

docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
fghaas: In the console dumps I have preferred to be as close to the real thing as possible.

docs/How-To_Guides/Object_storage/s3-credentials.md
evrardjp: It might not be obvious for some readers. Maybe adding `-id` would make it more obvious.

docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
evrardjp: I prefer it too.
I am just saying that because you don't write _fra1_ before, it might come as a surprise if the result is different.

But maybe I am too careful.

docs/How-To_Guides/Object_storage/sse-c.md
fghaas: @noonedeadpunk Good catch, thanks!

docs/How-To_Guides/Object_storage/sse-c.md
```diff
@@ -0,0 +1,90 @@
...
+3. Create a configuration file named `~/.rclone.conf`, with the
+   following content:
+
+        [{{extra.brand|lower|replace(' ','')}}]
+        type = s3
+        provider = Ceph
+        env_auth = false
+        access_key_id = <access key>
+        secret_access_key = <secret key>
+        endpoint = <region>.citycloud.com:8080
```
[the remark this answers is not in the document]
fghaas: Thanks again.

docs/How-To_Guides/OpenStack/Barbican/index.md
fghaas: Well if you're not `root`, then `pip install` will fail anyway. Except if you're in a virtualenv, in which case it's perfectly appropriate. No?

docs/How-To_Guides/Object_storage/sse-c.md
fghaas: Fixed.

docs/How-To_Guides/OpenStack/Barbican/share-secret.md
fghaas: Is that list meant to be permanent/authoritative?

docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
fghaas: The OpenStack CLI console dumps are a real issue, docs-wise, because they're incredibly wide. They're almost impossible to include even in the web output (without forcing the reader to scroll horizontally), and in portrait-mode PDF there isn't any chance at all to include them in a meaningful way. Other output formats besides the default `table` are hardly any better. `value` is useless (doesn't show names), `shell` and `json` are practically unreadable, `yaml` is... well, YAML.

evrardjp: I agree with you on all your points.
I am just saying we didn't start by saying that "For this example, we'll use the Fra1 region".

That's all.

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
fghaas: That sounds to me like it should be in a separate bit of documentation dealing with live migration and maintenance.

docs/How-To_Guides/Object_storage/s3-credentials.md
fghaas: Yeah, fair point. Official S3 documentation also appears to talk about "access key IDs" across the board. Fixed.

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
fghaas: Yes, in a separate PR.

docs/How-To_Guides/OpenStack/Barbican/generic-secret.md
fghaas: Fixed in latest version.

docs/How-To_Guides/OpenStack/Barbican/index.md
evrardjp: It works as a non-root user.

docs/How-To_Guides/OpenStack/Cinder/encrypted-volumes.md
evrardjp: seems fine to me. @noonedeadpunk  do you propose a new PR on top?

docs/How-To_Guides/OpenStack/Barbican/index.md
fghaas: Correcting myself:

* Prior to pip 20.0.0, running `pip install` as non-`root` would fail because it would attempt to write to an unwritable location.
* [Since pip 20.0.0](https://pip.pypa.io/en/stable/news/#id434), invoking just `pip` as non-`root` is equivalent to invoking `pip --user`.

Which means that

* If you are `root` and want to `pip`-install a package globally, the command as shown is correct.
* If you are not `root` and want to install a package in the user site directory, as of pip 20.0.0 (2+ years ago), the command as shown is correct.
* Regardless of whether you are `root` or not, and you're in a virtualenv and want to `pip`-install a package into that virtualenv, the command as shown is correct.

Thus, I'd prefer to leave it as-is.

fghaas:
I *think* all the open issues in this PR have been addressed (except the ones we deferred to a separate PR), is this good to merge now?

docs/How-To_Guides/OpenStack/Barbican/index.md
evrardjp: Fine for me. 
I guess I am old and habits are hard to break.

docs/How-To_Guides/OpenStack/Barbican/share-secret.md
evrardjp: Yes. definitive.
It can evolve over time, but it is still definitive.

evrardjp (approved):
You know best.

PR #12, opened 2022-03-16
build: Pass positional arguments into "tox -e serve"

fghaas:
It's usually helpful to be able to set a non-standard `-a` option to `mkdocs serve`, in order to listen on a different port if port 8000 is
already in use. Thus, pass positional arguments into the `serve` testenv.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,7 +20,7 @@ commands =
 
 [testenv:serve]
 commands =
-  mkdocs serve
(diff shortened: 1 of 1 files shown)


evrardjp:
I meant I didn't test locally.
The CI tests seem to run fine.

Sometimes GH actions take time to get a worker, it's normal, especially at the end of day in europe.

PR #13, opened 2022-04-19
Cleura branding updates

fghaas:
With the move from City Network to Cleura, update the branding:

* Update the color palette.
* Add the Cleura logo and favicon.
* Update City Network and City Cloud references to Cleura and Cleura Cloud.

Also, add a facility to deploy rendered content to GitHub pages (so that this can go live on the target site, via a CNAME update).

Fixes #2.

PR #14, opened 2022-04-19
build: Revamp and future-proof navigation

fghaas:
This is a PR to land on top of #13.

MkDocs has a rather tedious navigation management facility by default: either one defines the full navigation structure in the `nav` map in `mkdocs.yml`, or one omits `nav` altogether and opts for the fully autogenerated documentation structure, which does not provide nearly enough flexibility.

The middle ground is using the `awesome-pages` plugin, which autogenerates the site structure, but also allows us to customize the navigation when needed (and only when needed), by dropping `.pages` files into the relevant directories.

This also allows us to have mixed-case section names (like "OpenStack" or "How-To Guides") while keeping directory names (and hence URLs) all-lowercase.


evrardjp:
Ok, I think I get the gist of that addition. 

I am not sure it's a good thing to do:
- Adding a dependency (to maintain)
- Adding a bunch of files which are now required to write (without an explanation on how contributions need to happen in the future, though it's obvious when checking these commits)

In other words, "Best" can be the enemy of "good". This looks best, not sure if it is good in the long run.

However, I see no reason to block this initiative: Your team is tackling the two pain points I mention above, so I trust your best judgement.

docs/howto/openstack/octavia/tls-lb.md
```diff
@@ -1,10 +1,10 @@
 # HTTPS-terminating load balancers
 
 In {{extra.brand}}’s load balancing service, [OpenStack
-Octavia](https://docs.openstack.org/octavia), you can configure load
-balancers so that they manage HTTPS termination. That is to say that
-the load balancer encrypts and decrypts HTTPS traffic, and forwards
-HTTP to and from a backend web server.
+Octavia](https://docs.openstack.org/octavia/latest/), you can
```
evrardjp, line 4: shouldn't this fix be in another PR?

evrardjp (approved):
Ok for the change, please comment on whether you will split the unrelated link change in this PR (as this PR is refactoring links anyway).

PR #15, opened 2022-04-20
Upgrade the site from "experimental" to "beta", disable blank content, simplify deployment

fghaas:
As we're getting ready to publish this content, throw in a few updates to better reflect its current state:

* It's no longer "experimental" (a moniker that implies that the content might go away any moment), but it's far from complete. So call it "beta".
* Simplify the deployment workflow: rather than relying on a heavyweight external GitHub action, just use `mkdocs gh-deploy` (via a tox testenv).
* Comment out some content that isn't yet ready for general consumption, so it isn't visible in the rendered web site. But leave it in-tree so that interested partners can dive into it.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+exec tox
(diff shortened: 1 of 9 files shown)


evrardjp (approved):
For the gh actions, that is by far closer to what I am used to see.
I like it.

I am really not fond of the massive PRs doing many things together. It's easy to build multiple PRs...

But anyway, let's not bicker too much. This is a good progress, let's go ahead with it.

PR #16, opened 2022-04-21
feat: Enable light and dark mode, and detect system setting

fghaas:
Being able to switch between light and dark mode is not a gimmick; it's an accessibility issue: people living with migraines often find it extremely uncomfortable and/or painful to look at large bright white areas on their screen.

Thus, add a light/dark mode switch, and also detect the global system or browser setting to provide a reasonable default.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -46,4 +46,13 @@ theme:
   logo: assets/logo.png
   name: material
   palette:
-    scheme: cleura
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
+      toggle:
+        icon: material/lightbulb-outline
+        name: "Switch to dark mode"
+    - media: "(prefers-color-scheme: dark)"
+      scheme: slate
+      toggle:
+        icon: material/lightbulb
+        name: "Switch to light mode"
--- a/theme/assets/stylesheets/cleura-palette.css
+++ b/theme/assets/stylesheets/cleura-palette.css
@@ -1,4 +1,9 @@
-[data-md-color-scheme="cleura"] {
+[data-md-color-scheme="default"] {
+    --md-primary-fg-color: #FF6600;
+    --md-accent-fg-color:  #2643FF;
+}
+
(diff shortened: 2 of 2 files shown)


mkdocs.yml
```diff
@@ -46,4 +46,13 @@ theme:
   logo: assets/logo.png
   name: material
   palette:
-    scheme: cleura
+    - media: "(prefers-color-scheme: light)"
+      scheme: default
```
evrardjp, line 50: nit: default may make us think it's a default color-scheme, where the extra_css in fact overrides that.
Would it make sense to call it cleura-light-mode and cleura-dark-mode ? For me, it looks more clear for marketing and other contributions.

evrardjp (commented):
The default is indeed adapting to your system, and flips to dark theme if that's what you prefer. So that's good to me.

Next to that, everything is readable, which is good enough for me.

I will wait for your comment on palette scheme name to vote.

mkdocs.yml
fghaas: No, this is actually me having misunderstood how color schemes are named in mkdocs-material, initially. The light scheme is properly called `default`, and the dark scheme is called `slate`. What you can then do is override individual colors in either scheme. 

If you create a new scheme it inherits from `default`. So if there's a theme called `cleura` the two colors from `default` are easy to override, but if you want to have `cleura-dark` that's actually `slate` with a few tweaks, you have to reimplement all colors from `slate` in your dark scheme. 

The toggle is now implemented as suggested in [the documentation](https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-palette-toggle), so as to follow [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

evrardjp: I guess I misunderstood https://squidfunk.github.io/mkdocs-material/setup/changing-the-colors/#color-scheme . 
Thanks for clarifications.

evrardjp (approved):

PR #17, opened 2022-04-21
docs: Update README

fghaas:
The site has moved on from an "experimental" to a "beta" state. Update the README accordingly, and add a call to action for anyone interested in contributing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,12 @@
-# Experimental documentation repository
+# Cleura Docs Beta Repository
 
-For the time being, please do not mistake this for anything official
-or supported.
+Here’s where the content of the <https://docs.cleura.cloud>
+documentation web site lives.
+
+This site is currently in a Beta state: we do know and understand that
+there is a lot of content that's still missing (in fact, we're just
+getting started) — but the content which is already there should be
+usable.
+
+If you’d like to help us build this, please take a look at the
+[Contribution Guide](https://docs.cleura.cloud/contrib/)!


README.md
```diff
@@ -1,4 +1,12 @@
-# Experimental documentation repository
+# Cleura Docs Beta Repository
 
-For the time being, please do not mistake this for anything official
-or supported.
+Here’s where the content of the <https://docs.cleura.cloud>
+documentation web site lives.
+
+This site currently in a Beta state: we do know and understand that
```
mrtmm, line 6: I think it's missing an 'is' here :)

fghaas: Ah-hah! Good catch, eagle-eye :)

fghaas: Fixed.

evrardjp:
Not sure why invalid url - https://docs.openstack.org/octavia/latest/  .

evrardjp:
Ok, force push landed now, will review when the build is passing.

fghaas:
This is so weird. The link checker chokes on <https://docs.openstack.org/octavia/latest/> about half of the time it checks, and the remedy is to just re-run the job.

evrardjp:
For what it's worth, when I was building other static sites with 404 checkers I often got that issue due to rate limiting and other issues. The first step would be to record the HTTP error code, and only then _maybe_ check this with opendev infra.

evrardjp (approved):

PR #18, opened 2022-04-22
build: Add CNAME file

fghaas:
GitHub Pages custom domains are meant to be enabled in one of two ways: via the repository settings (Settings → Code and automation → Pages → Custom domain), or via a `CNAME` file in the root of the repository's `gh-pages` branch.

It appears that the former option is not working, in the sense that it gets cleared on every deployment run. This, then, results in the site effectively disappearing (i.e. producing HTTP 404s).

Thus, opt for the latter option and do use the `CNAME` file.

### DIFF
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+docs.cleura.cloud


NamrataSitlani (approved):

fghaas:
Note to reviewers: the `docs/CNAME` file becomes `site/CNAME` upon `mkdocs build`, which then becomes `CNAME` in the root of the `gh-pages` branch upon `mkdocs gh-deploy`.

evrardjp:
Thanks for that note, because I was pretty sure CNAME had to be in gh-pages root.

evrardjp:
It only disappears in the system because that CNAME file is disappearing from your run, I think. But let's ignore that for a moment and let's get the job done.

evrardjp (approved):

fghaas:
> It only disappears in the system because that CNAME file is disappearing from your run, I think. But let's ignore that for a moment and let's get the job done.

Ah yes, you are correct. It looks like indeed what the modification to the settings is doing *is* just inject the `CNAME` file. At least that's how I read [this commit](https://github.com/citynetwork/docs/commit/22fed5a958013ef1629eb289b513d26f7b7b770a). (I didn't push anything at the time, but I did change that setting.)

fghaas:
No interruption on this deployment; settings are now unchanged, as expected. Thanks for the prompt review.

PR #24, opened 2022-08-31
fix: IPv6 in compliant

packettoobig:
sto1hs and sto2hs have no IPv6 support yet
this work is not prioritized as of today


fghaas (approved):
News to me! Not sure how this slipped through the initial review. Thank you!

PR #27, opened 2022-09-14
feat: Add Kubernetes cluster management to feature support matrix

fghaas:
Since the closed beta for Gardener in Public Cloud is now underway, mention it as a feature that is in deployment, but not yet supported.

Simultaneously, also add Magnum to the feature matrix, which has been officially supported since 2018.

Reference:
https://cleura.com/press-releases/city-network-launches-container-service-in-city-cloud-with-kubernetes-in-focus/

### DIFF
--- a/docs/reference/features/public.md
+++ b/docs/reference/features/public.md
@@ -51,3 +51,9 @@
 | Application layer (HTTPS, with secrets management for TLS certificates) | :material-check: | :material-check: | :material-check: | :material-check: | :material-check: |
 
 
+## Kubernetes management
+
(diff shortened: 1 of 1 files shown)


mrtmm (dismissed):
Looks good to me, thanks!

martindahlswe (dismissed):
Looks good.

fghaas:
Turns out that this wasn't accurate after all; the closed beta only encompasses the European Cleura Cloud regions. @mrtmm, can I have a re-approval please?

mrtmm (approved):

PR #29, opened 2022-10-10
feat: Expand S3 object versioning how-to

fghaas:
Explain how to list the versions available for an object, how to create a versioned object, and how to remove the latest version or a specific version of an object.

Also, apply some whitespace fixes.

### DIFF
--- a/docs/howto/object-storage/s3/versioning.md
+++ b/docs/howto/object-storage/s3/versioning.md
@@ -12,8 +12,8 @@ To enable versioning in a bucket, use one of the following commands:
     ```bash
     aws --profile <region> \
       --endpoint-url=https://s3-<region>.{{extra.brand_domain}}:8080 \
-      s3api put-bucket-versioning \ 
-      --versioning-configuration Status=Enabled \ 
+      s3api put-bucket-versioning \
+      --versioning-configuration Status=Enabled \
       --bucket <bucket-name>
     ```
 === "mc"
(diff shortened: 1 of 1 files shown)


colder-is-better:
It all looks good to me, but allow me to point out a pico-typo and suggest some minor changes to the syntax which contribute to an infinitesimally better text -- or at least that's what I think.

From this sentence...

If the object does already exist, it is not replaced. Instead, a new version appears, in addition to the old one (**BEFORE**)

...I would delete the last comma:

* If the object does already exist, it is not replaced. Instead, a new version appears in addition to the old one (**AFTER**) 

The "If" in this one...

If a bucket that has versioning enabled, you may list the versions available for an object: (**BEFORE**)

...is clearly an "In" (or we can leave the "If" and delete the first "that"):

In a bucket that has versioning enabled, you may list the versions available for an object: (**AFTER**)

In this paragraph...

When you download an object from a versioned bucket *without* specifying a version identifier, your S3 client will download the latest version of the object that the bucket contains (**BEFORE**)

...I would delete the final "that":

When you download an object from a versioned bucket *without* specifying a version identifier, your S3 client will download the latest version of the object the bucket contains (**AFTER**)

I would rephrase this paragraph...

Like the commands to *create* objects, the commands to *delete* them behave differently once object versioning is enabled on a bucket (**BEFORE**)

...and make it like this:

Like the commands for object *creation*, the commands for object *deletion* behave differently once object versioning is enabled on a bucket (**AFTER**)

Finally, I would slightly change this sentence... 

The command to delete an object will not delete the object altogether, but instead revert it to the prior version (**BEFORE**)

...and make it like so (no "instead"):

The command to delete an object will not delete it altogether but revert it to the prior version (**AFTER**)

docs/howto/object-storage/s3/versioning.md
```diff
@@ -44,6 +44,61 @@ the following commands:
...
+    ```
+=== "s3cmd"
+    ```bash
+    s3cmd put <local-filename> s3://<bucket>
+    ```
+
+
+## Listing object versions
+
+If a bucket that has versioning enabled, you may list the versions
```
colder-is-better, line 81: Minor typo, this sentence may start like "In a bucket that has versioning" or like "If a bucket has versioning".

docs/howto/object-storage/s3/versioning.md
```diff
@@ -53,22 +108,72 @@ the following commands:
...
+      --version-id <versionid> \
+      <region>/<bucket-name>/<object-name> \
+      <local-filename>
     ```
 === "s3cmd"
     This functionality is not available with the `s3cmd` command.
 
 When you download an object from a versioned bucket *without*
 specifying a version identifier, your S3 client will download the
 latest version of the object that the bucket contains.
```
colder-is-better, line 129: I would delete the last "that" in the sentence: "that latest version of the object the bucket contains".

docs/howto/object-storage/s3/versioning.md
```diff
@@ -53,22 +108,72 @@ the following commands:
...
+
+
+## Deleting a versioned object
+
+Like the commands to *create* objects, the commands to *delete* them
```
colder-is-better, line 134: I would slightly rephrase: "Like the commands for object creation, the commands for object deletion behave differently"

docs/howto/object-storage/s3/versioning.md
```diff
@@ -53,22 +108,72 @@ the following commands:
...
+behave differently once object versioning is enabled on a bucket.
+
+The command to delete an object will not delete the object altogether,
+but instead revert it to the prior version. The exception to this rule
```
colder-is-better, line 138: I suggest a slight rephrasing: "The command to delete an object will not delete it altogether but revert it to the prior version"

docs/howto/object-storage/s3/versioning.md
```diff
@@ -44,6 +44,61 @@ the following commands:
...
+## Creating a versioned object
+
+Once object versioning is enabled on a bucket, the normal object
+creation and replacement commands behave in a manner different from
+that in unversioned buckets:
+
+* If the object does not already exist, it is created (as in an
+  unversioned bucket).
+* If the object does already exist, it is not replaced. Instead, a new
+  version appears, in addition to the old one.
```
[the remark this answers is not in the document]
fghaas: Yes.

docs/howto/object-storage/s3/versioning.md
fghaas: I meant "in". Thanks for spotting. :)

docs/howto/object-storage/s3/versioning.md
fghaas: Better still, trim the sentence a bit. "... the latest version of that object."

docs/howto/object-storage/s3/versioning.md
fghaas: I would disagree. :) When given the choice to describe an action with a verb or with an action noun, the verb is the superior option because it is more concise and direct.

docs/howto/object-storage/s3/versioning.md
fghaas: Good suggestion, and we can trim it even further by removing the word "altogether".

fghaas:
@Antek74 Are you OK with this update landing in its current form?

fghaas:
@colder-is-better We haven't had any objections on this one, so I've rebased it on current `main` and would now need one more review and approval so we can merge this.

colder-is-better (approved):

PR #30, opened 2022-10-11
fix: Update version references in Compliant Cloud

fghaas:
As of October 2022, all our Compliant Cloud deployments run Ceph Pacific and OpenStack Xena. Update the version matrix accordingly.


fghaas:
@Antek74, any objections to merging this?

PR #31, opened 2022-10-12
Change and use capital letters on public regions name

ahosam:
unified the regions naming by using capital characters

### DIFF
--- a/docs/reference/versions/public.md
+++ b/docs/reference/versions/public.md
@@ -2,7 +2,7 @@
 
 ## OpenStack Services
 
-|                                | Kna1  | Sto2   | Fra1  | Dx1   | Tky1   |
+|                                | KNA1  | STO2   | FRA1  | DX1   | TKY1   |
 | ------------------------------ | ----- | ------ | ----- | ----- | ------ |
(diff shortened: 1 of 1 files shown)


fghaas:
@ahosam, please take a moment to review [this blog post](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html), and then improve your commit message based on that. The commit message shouldn't explain *what* the change does, but *why* the change is necessary and why this way is the best to address that necessity. Note also that we use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) so the commit message subject would need a `fix:` prefix.

However, all that aside, are you sure that the change as such makes sense? Our region names don't use all-caps in the Keystone catalog, and [cleura.cloud](https://cleura.cloud) also doesn't capitalize the region names. What's the rationale for using all-caps in the docs?

ahosam:
Ack

ahosam:
I checked the keystone catalog for all public regions, Yes, it is not make sense to change it on docs, will close this PR and reconsider your comments :) for the next

fghaas:
OK, that works for me! :)

PR #36, opened 2022-10-20
feat: Add HowTo re creating a network from web GUI

colder-is-better:
We walk through creating a new full-featured network using the Cleura Cloud Control Panel.


fghaas (changes requested):
This PR needs a rewrite, please. The Cleura Cloud Control Panel is **not** OpenStack Horizon, and we must not confuse the two. So you'll need to remove the Horizon references, and also rename the Markdown file (thereby also changing the URL it gets rendered to). 

Please fix that first, and then we can proceed with a line-by-line review. Thanks!

colder-is-better:
> This PR needs a rewrite, please. The Cleura Cloud Control Panel is **not** OpenStack Horizon, and we must not confuse the two. So you'll need to remove the Horizon references, and also rename the Markdown file (thereby also changing the URL it gets rendered to).
>

It is surely not, but I imagined we can still use some official OpenStack terminology.
 
> Please fix that first, and then we can proceed with a line-by-line review. Thanks!

Done :)

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,58 @@
...
+Before creating a server in {{extra.brand}} Cloud, you need to have a 
+network to make the new server a member of. You may have more than one 
+network per region, so let us now walk through creating a new network 
+using the {{extra.brand}} Cloud Control Panel. Fire up your favorite web 
+browser, navigate to the [Cleura Cloud](https://cleura.cloud) page, and 
+login into your {{extra.brand}} account.
+
+On the top right-hand side of the {{extra.brand}} Cloud Control Panel, 
+click the _Create_ button. A new pane will slide into from the 
+right-hand side of the browser window, titled ---unsurprisingly
```
fghaas, line 12: Please remove the "unsurprisingly enough" bit. In general, we want to avoid phrases that sound like something is obvious or excessively easy, because obviousness and ease are quite subjective — and such phrases might come across as condescending to people who find something not so obvious and easy.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,58 @@
...
+You will notice several rounded boxes prominently displayed on that 
+pane, each for defining, configuring, and instantiating a different 
+{{extra.brand}} Cloud object. Go ahead and click the _Network_ box. A 
+new pane titled _Create Network_ will slide over. At the top, type in a 
+name and select one of the available regions for the new network.
+
+![New network name and region](assets/new-net-panel/shot-02.png)
+
+Expand the _Advanced Options_ section below and make sure _Port 
+Security_ is enabled. You do not have to set a value for the MTU
```
fghaas, line 26: I would suggest to say "Leave the MTU parameter blank". That way the phrasing is affirmative rather than negative, and the recommendation is stronger than saying "you don't have to" (which implies "... but you can", and in most circumstances people really should *not* override the default MTU size).

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,58 @@
...
+![New network name and region](assets/new-net-panel/shot-02.png)
+
+Expand the _Advanced Options_ section below and make sure _Port 
+Security_ is enabled. You do not have to set a value for the MTU 
+parameter.
+
+![MTU and port security](assets/new-net-panel/shot-03.png)
+
+You probably want a full-featured network for your cloud servers, so 
+unless you know what you are doing, please activate the _Create a
```
fghaas, line 32: Remove the "so unless you know what you are doing, please" bit.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,58 @@
...
+Expand the _Advanced Options_ section below and make sure _Port 
+Security_ is enabled. You do not have to set a value for the MTU 
+parameter.
+
+![MTU and port security](assets/new-net-panel/shot-03.png)
+
+You probably want a full-featured network for your cloud servers, so 
+unless you know what you are doing, please activate the _Create a 
+complete network containing a subnet and a router_ option. You will 
+notice that a CIDR is pre-configured for your network. You also get a
```
fghaas, line 34: I know this isn't your fault but the GUI's (and Neutron's), but "a CIDR" is "a Classless Inter-Domain Routing", which doesn't make sense. :) 

I would put this:

```markdown
You will notice that 
[a network address in CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation)
is pre-configured for your network.
```

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,58 @@
...
+using the {{extra.brand}} Cloud Control Panel. Fire up your favorite web 
+browser, navigate to the [Cleura Cloud](https://cleura.cloud) page, and 
+login into your {{extra.brand}} account.
+
+On the top right-hand side of the {{extra.brand}} Cloud Control Panel, 
+click the _Create_ button. A new pane will slide into from the 
+right-hand side of the browser window, titled ---unsurprisingly 
+enough--- _Create_.
+
+![Create a new object](assets/new-net-panel/shot-01.png)
```
fghaas, line 15: In the screenshots, you will want to blur the project name (in the GUI element labeled _Project,_ at the top of the screen).

fghaas (commented):
Looks good overall! Just a few suggestions.

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: Sure, removed it. That was a joke, actually, one that I was pretty sure would not survive the final version of the how-to. But I included it anyways :)

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: Much better to follow your approach, yes.

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: Removed it, but this phrase in general doesn't hurt IMHO.

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: Rephrased per your suggestion. (BTW, I know what a CIDR notation is but, yeah, my phrasing was poor.)

fghaas (commented):
This looks good to me know. Thanks!

@murwall or @Skipper126, if you have any objections to this being merged, please holler. Otherwise, we'll merge this in a few hours.

docs/howto/openstack/neutron/new-network-control-panel.md
fghaas: I think it really does. "Unless" implies a negative. If I say "unless you have a death wish, you shouldn't ride your bicycle in rush-hour traffic in Athens" implies that the reasonable assumption is that you _don't_ have a death wish. 

Writing "unless you know what you are doing" thus implies that we default to the assumption that the reader does not know what they're doing, and that's a rude assumption.

Skipper126:
Hey! Thanks for contributing, amazing work. :) Content wise looks great. I'll probably have some comments regarding usage of config values interpolation. I'll put them shortly, so please hold on.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,57 @@
+# Creating a new network from the Cleura Cloud Control Panel
+
+Before creating a server in {{extra.brand}} Cloud, you need to have a 
+network to make the new server a member of. You may have more than one 
+network per region, so let us now walk through creating a new network 
+using the {{extra.brand}} Cloud Control Panel. Fire up your favorite web
```
Skipper126, line 6: We can use {{config.extra.gui}} value here instead, which is "Cleura Cloud Control Panel".

Just a heads-up: For some reason the config variables don't work when used in page title, so leave line 1 as it is. :)

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -0,0 +1,57 @@
+# Creating a new network from the Cleura Cloud Control Panel
+
+Before creating a server in {{extra.brand}} Cloud, you need to have a 
+network to make the new server a member of. You may have more than one 
+network per region, so let us now walk through creating a new network 
+using the {{extra.brand}} Cloud Control Panel. Fire up your favorite web 
+browser, navigate to the [Cleura Cloud](https://cleura.cloud) page, and
```
Skipper126, line 7: This is up for a discussion, but in another docs page, file create-shoot-cluster.md line 13, when including cleura.cloud link, we use `<https://{{config.extra.gui_domain}}>` instead . This is later rendered as a full link on a page.
Both ways are fine, but for consistency reasons, I think it's better to stick with one approach.

fghaas:
@Skipper126 is correct about using config variable substitution for those variables which we already define. I haven taken the liberty to amend the commit to that effect.

fghaas (approved):

PR #37, opened 2022-10-21
chore: Add .gitignore file

colder-is-better:
There are specific MacOS-generated files we always want to ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.DS_Store
+
+!/.gitignore


.gitignore
```diff
@@ -0,0 +1,5 @@
+.DS_Store
+.tox/*
+site/*
```
fghaas, line 3: Please, let's not add `.tox` and `site` to `.gitignore`. I am quite fond of the idea of being able to restore a Git checkout to its pristine condition with `git clean -fd`. :)

colder-is-better: Aha, I understand!

fghaas:
This looks good now but the commit message still needs a [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) prefix (in this case, `chore:`).

fghaas (approved):
Thank you!

PR #38, opened 2022-10-21
WIP: feat: Extend new network how-to for OpenStack CLI

fghaas:
Rather than having separate pages for the Control Panel and CLI which both cover new network, we can have one page that shows the required steps side-by-side for both, using content tabs.

Reference:
https://squidfunk.github.io/mkdocs-material/reference/content-tabs/


fghaas:
This WIP PR has been superseded by #44 and won't be merged.

PR #39, opened 2022-10-25
feat/refactor: Expand Git information in contribution guide, split into 3 pages

fghaas:
In order to make contributions to the documentation more accessible to
occasional or casual contributors, spell out how to fork and clone the
documentation repo in the contribution guide, using two approaches:

1. Using the GitHub web interface plus the `git` CLI,
2. Combining the `git` and `gh` CLIs.

This then makes the single-page configuration guide a little unwieldy,
so split it into 3 pages.

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -1,0 +1,88 @@
+# Modifying content on this site
+
+You have two options for editing content: directly in your browser
+using GitHub, or using a Git-based workflow from your local work
+environment.
+
+
+## Modifying content from your browser
+
+Every page on this site has an Edit button 🖍️. If you click it, it’ll
+take you straight to the corresponding source page in GitHub. Then,
+you can follow [GitHub’s
+documentation](https://docs.github.com/en/repositories/working-with-files/managing-files/editing-files#editing-files-in-another-users-repository)
+on how to propose changes to another user’s repository.
+
+
+## Modifying using Git
+
+The Git repository for this site lives at <{{ config.repo_url }}>. You
+can [fork that
+repository](https://docs.github.com/en/get-started/quickstart/fork-a-repo),
+make the proposed changes in your fork, and then send us a standard
+[GitHub pull
+request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/about-pull-requests).
+
+For this purpose, use `git` in combination with either GitHub's web
+interface, or the `gh` command-line interface (CLI).
+
+First, create a fork of the documentation repository:
+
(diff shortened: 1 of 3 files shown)


docs/contrib/modifications.md
```diff
@@ -0,0 +1,88 @@
...
+    Then, open your browser to the URL suggested by the `git push`
+    command, and proceed to create a pull request.
+=== "`gh` client"
+    ```bash
+    gh pr create --fill
+    ```
+
+## Monitoring changes as you edit
+
+If you would like to see your changes as you are working on them, you
```
colder-is-better, line 75: I think it would be better (simpler) to replace "If you would like" with "If you want". Also, grammatically it is better to replace "as you are working" with "as you work". So, overall, it is better to start like this: "If you want to see your changes as you work on them...".

docs/contrib/quality.md
```diff
@@ -0,0 +1,38 @@
+# Quality checks
+
+There are a few checks that we apply to the configuration of this
```
colder-is-better, line 3: I would delete the "that", so the sentence would read like this: "There are a few checks we apply to the configuration..."

docs/contrib/quality.md
```diff
@@ -0,0 +1,38 @@
...
+* We check the commit message with
+  [gitlint](https://jorisroovers.com/gitlint/), and enforce the
+  [Conventional
+  Commits](https://www.conventionalcommits.org/en/v1.0.0/) commit
+  message style.
+* We check whether the documentation still builds correctly, with your
+  change applied.
+* We check to make sure that no internal or external links in the
+  documentation are dead. This is one example where the checks might
+  fail through no fault of yours at all — some external link may have
```
colder-is-better, line 16: The "at all" is not needed, I would even say that it is even wrong. So, in my opinion, the sentence should start like this: "This is one example where the checks might fail through no fault of yours...".

colder-is-better (commented):
All in all, it looks good to me. I have only some femto-nits regarding grammar and/or style.

docs/contrib/quality.md
fghaas: I agree.

docs/contrib/quality.md
fghaas: That is indeed more elegant, but I think that including "that" (or, where appropriate, replacing it with ", which") improves the reading flow for non-native speakers. I would prefer to leave this as-is.

fghaas:
Thanks! I've update the PR according to your suggestions, please approve if you're OK with this being merged.

colder-is-better (approved):
Looks more than good to me now, thanks!

PR #43, opened 2022-10-26
fix: Replace IPv4 addresses with RFC5737

packettoobig:
I believe it make sense to avoid using real IPs in our documentation. So I would suggest using IP blocks that are actually reserved for this use by IANA.

See [RFC5737](https://www.rfc-editor.org/rfc/rfc5737.html) section 3.


fghaas:
Looks good, thanks! I'll go ahead and fix the commit message so that `gitlint` stops complaining.

PR #44, opened 2022-10-30
refactor: Add OSC approach and employ content tabs

colder-is-better:
We show how to create networks in Cleura Cloud using two approaches: one of the Control Panel and one of the OpenStack CLI.

To demonstrate each way of working, we present two specific examples in two different content tabs.


fghaas:
I'm sorry, but as-is this I find myself unable to do a meaningful review for this PR. :( I am about to force-push to your branch with the following changes:

* Separate the file modifications and the rename into two separate commits (to be squashed later).
* Wrap the excessively long lines.
* Rebase on current `main`.

That way, we'll get a meaningful diff between the current state of `main` and the proposed change.

fghaas:
@colder-is-better For me, the page now renders correctly, even with wrapped lines. I'll now go ahead and start with a proper content review.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
-# Creating a new network from the Cleura Cloud Control Panel
+# Creating new networks on the Cleura Cloud
```
fghaas: Why not simply "Creating a new network"?

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
-new pane titled _Create Network_ will slide over. At the top, type in a 
-name and select one of the available regions for the new network.
+## Creating a network
 
-![New network name and region](assets/new-net-panel/shot-02.png)
+To create a network from the {{extra.gui}}, fire up your favorite web
+browser, navigate to the [Cleura Cloud](https://{{extra.gui_domain}})
+page, and login into your {{extra.brand}} account. On the other hand,
+if you prefer to work with OpenStack CLI, please do not forget to
+source the relative RC file first.
```
fghaas: Don't put this outside the content tab. One of these clearly applies to the GUI, the other to the CLI. They should be in their respective content tabs.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    pane, each for defining, configuring, and instantiating a different
+    {{extra.brand}} Cloud object. Go ahead and click the _Network_ box. A
+    new pane titled _Create Network_ will slide over. At the top, type
+    in a name and select one of the available regions for the new
+    network.
 
-![Complete network](assets/new-net-panel/shot-04.png)
+    ![New network name and region](assets/new-net/shot-02.png)
+=== "OpenStack CLI"
+    Start by creating a new network, named `nuremberg`:
```
fghaas: Please don't call the network `nuremberg`. People might read this as a [dog whistle](https://en.wikipedia.org/wiki/Dog_whistle_(politics)) reference to Nazi party rallies.

The GUI examples use the innocuous name `nordostbahnhof` for the network, so in the interest of parallelism, use that here, too.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    the new network will be readily available.
+
+    ![Finish creating network](assets/new-net/shot-05.png)
+=== "OpenStack CLI"
+    You now have to create a subnet for the new network. Let us call
+    this subnet `nuremberg_subnet`:
+
+    ```bash
+    openstack subnet create nuremberg-subnet \
+        --network nuremberg --subnet-range 10.20.30.0/24
```
fghaas: `s/nuremberg/nordostbahnhof/g` please.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    ```bash
+    ROUTER_ID="566de991-fc0e-4f85-b6c4-5c87694781f7"
+    ```
+
+    You want the `nuremberg_router` connected to the external
+    network. To realize this connection, you need the `id` of said
+    network. Go ahead and list all available networks...
+
+    ```bash
+    openstack network list -f json
```
fghaas: Please don't use `-f json`. When we create objects in OpenStack, the output we get is tabular (that is, it is equivalent to the default format, i.e. `-f table`). Instructing users to then read back the object data in a different format is a [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) violation.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    | shared                    | False                                |
+    | status                    | ACTIVE                               |
+    | subnets                   |                                      |
+    | tags                      |                                      |
+    | updated_at                | 2022-10-30T14:31:50Z                 |
+    +---------------------------+--------------------------------------+
+    ```
 
-![All networks in all regions](assets/new-net-panel/shot-06.png)
+## Adding subnet, router, and finishing up
```
fghaas: "Finishing up" sounds a bit odd here. "Adding a subnet and router" would be better, I think.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    corresponding three-dot icon (right-hand side) and select _View
+    details_. Then, you can glance over all the details regarding the
+    selected network's ports, subnets, and routers.
+
+    ![Network details](assets/new-net/shot-07.png)
+=== "OpenStack CLI"
+    To list all available networks in a specific region, just type:
+
+    ```bash
+    openstack network list -f json
```
fghaas: Drop `-f json`.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    ```
+
+    You can always ask for more specific results. For instance, to see
+    all internal networks only, type the following:
+
+    ```bash
+    openstack network list --internal -f json
```
fghaas: Drop `-f json`.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    ```
+
+    You can also get detailed information about a specific network:
+
+    ```bash
+    openstack network list --name nuremberg --long -f json
```
fghaas: Drop `-f json`.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,332 @@
...
+    ```
+
+    ```plain
```
fghaas: This JSON output shouldn't be here, but whenever you do want to include JSON output, the proper syntax highlighting hint would be `json`, not `plain`.

fghaas (commented):
I have a few comments.

I'll also force-push another fixup that temporarily factors out the filename changes for the screenshots.

fghaas:
By the way, what you likely want to do now in your local checkout is

```bash
git checkout main
git pull origin main
git fetch <remote-pointing-to-your-fork>
git checkout net_creation_howto_ext
git reset --hard '@{u}'
git rebase -i --no-autosquash main
```
Then say `edit` for the one non-fixup commit, and close your editor.

Then make your local changes, and finally run

```bash
git commit -a --amend --no-edit
git rebase --continue # complete the rebase
git push --force-with-lease
```

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: I thought to use this paragraph as a "starter" for the whole section. Sort of saying something generic enough and useful, then spring into action.

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: My idea is to create two different networks, each using a different method. (I mention this in the opening paragraph, by the way.) I understand how `nuremberg` can be misunderstood, though. So, my suggestion is to do a `s/nuremberg/hauptbahnhof/g`. What do you think?

docs/howto/openstack/neutron/new-network-control-panel.md
colder-is-better: Aha, I never heard of POLA! The thing is that tabular output --at least on my terminal-- seems ugly -- and what I really want to say is "broken". That's why I went for JSON. Do you think we could ignore POLA, at least for this post?

docs/howto/openstack/neutron/new-network-control-panel.md
fghaas: No. You are documenting, in one page, how to create a network with the GUI, and with the CLI. Please don't switch up the names. Use one network name throughout.

docs/howto/openstack/neutron/new-network-control-panel.md
fghaas: You can never "ignore" POLA.

And, we use tabular output everywhere else. If you want to recommend switching to JSON in this case, you'd either have to change *all* `openstack <object> [show|list]` examples throughout,  you'd at the very least have to explain (to readers in the text, not to me in the review thread) why you do so. "The other format is ugly" is not a sufficient explanation, really. If the other format is *broken,* then we shouldn't use it anywhere, but I don't exactly see the brokenness. Can you elaborate?

colder-is-better: Well, this is how tabular output looks on my terminal (and that's why I went for JSON).

Screen Shot 2022-10-31 at 14 31 51

fghaas: In that case generate the terminal output with ``openstack network list --max-width 100``.

fghaas: See also: #46 (I guess this is quite a helpful tweak to suggest to users).

colder-is-better: Agreed, setting CLIFF_FIT_WIDTH solves any issues regarding the looks of tabular output. I am going ahead and making all the necessary edits.

fghaas:
The most recent change didn't follow the rebase instructions, and mixed in content changes with the last fixup commit. I'll force-push this one more time *without* all the file renames. Then we fix the content, then we rename the files.

fghaas (commented):
A few more notes.

docs/howto/openstack/neutron/new-network-control-panel.md
```diff
@@ -1,57 +1,250 @@
-# Creating a new network from the Cleura Cloud Control Panel
+# Creating new networks on the Cleura Cloud
```
fghaas, line 1: I suggest to replace "on the Cleura Cloud" with "in Cleura Cloud", which is also what the rest of the page uses.

fghaas (commented):
I think this is very nearly good to go now; I only have one more tiny suggestion about the title.

fghaas:
And you can rewrite the commit message subject to something like "feat: Add CLI to new network howto".

PR #45, opened 2022-10-31
build: Use environment variable for toggling htmlproofer

fghaas:
The htmlproofer plugin increases documentation build times nearly tenfold. This may be undesirable when making quick changes while running `tox -e serve`. Thus, introduce an environment variable, `DOCS_ENABLE_HTMLPROOFER`, and disable the plugin if it is set to `false`.[^1]

Since this means that `mkdocs.yml` no longer contains `true` or `false` for a boolean variable, disable the yamllint `truthy` check for that line.[^2]

Also, update the contribution guide to that effect.

[^1]: https://www.mkdocs.org/user-guide/configuration/#environment-variables
[^2]: https://yamllint.readthedocs.io/en/stable/disable_with_comments.html#disabling-checks-for-a-specific-line

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -85,4 +85,13 @@ tox -e serve
 A local copy of the documentation will then run on your local machine
 and be accessible from <http://localhost:8000> in your browser.
 
+When you are planning to make several changes in rapid succession, you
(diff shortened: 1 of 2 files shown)


colder-is-better (approved):
This is a very useful feature indeed -- and everything LGTM, by the way.

PR #48, opened 2022-10-31
build: Don't show email addresses for authors

fghaas:
Since version 0.6.5, the git-authors plugin has the option of only showing the author's name(s), rather than rendering them as a
`mailto:` link. Since we want to list contributors but don't want to open them to spam, set the `show_email_address` option to `false`.

Also, fix up the `.mailmap` file so author identities are properly merged, and every person shows up as an author only once for every page (in other words, we avoid listings like "John Doe, Emily Example, John Doe" under a page).

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,15 +1,12 @@
 # For the purpose of this file, see:
 # https://timvink.github.io/mkdocs-git-authors-plugin/mailmap.html
-Christos Varelas <christos.varelas@citynetwork.eu>
+Christos Varelas <christos.varelas@cleura.com> <christos.varelas@citynetwork.eu>
 Christos Varelas <christos.varelas@cleura.com>
-Dmitriy Rabotyagov <dmitriy.rabotyagov@citynetwork.eu>
-Dmitriy Rabotyagov <dmitriy.rabotyagov@cleura.com>
+Dmitriy Rabotyagov <dmitriy.rabotyagov@cleura.com> <dmitriy.rabotyagov@citynetwork.eu>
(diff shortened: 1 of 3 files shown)


.mailmap
```diff
@@ -1,15 +1,12 @@
...
-Florian Haas <florian@citynetwork.eu>
-Florian Haas <florian@cleura.com>
-Foad Lind <foad.lind@citynetwork.eu>
-Foad Lind <foad.lind@cleura.com>
-Mateusz Guziak <mateusz.guziak@cleura.com>
-Mateusz Guziak <skipper126@interia.pl>
-Namrata Sitlani <namrata.sitlani@citynetwork.eu>
-Namrata Sitlani <namrata.sitlani@cleura.com>
+Florian Haas <florian@cleura.com> <florian@citynetwork.eu>
+Florian Haas <florian@cleura.com> <florian.haas@cleura.com>
```
colder-is-better, line 8: Can you have a single line with one full name and three email addresses?

colder-is-better (commented):
Everything is looking good to me -- I have but one question!

.mailmap
fghaas: Nope, that's not how `.mailmap` works. See [the documentation](https://git-scm.com/docs/gitmailmap) (if you can parse the famously impenetrable git manpage language; see also <https://git-man-page-generator.lokaltog.net> for comic relief).

fghaas:
@colder-is-better if you mark your review with an approval (at your convenience), then this change will auto-merge.

colder-is-better (approved):
LGTM -- thanks!

PR #49, opened 2022-11-01
Add beta disclaimer and issue templates

fghaas:
Add a couple of patches to better explain what "beta" means for the site, and to enable people to file bugs if they see them:

- feat: Expand Beta disclaimer, add call to action for documentation bugs
- docs: Add GitHub issue templates

PR #50, opened 2022-11-01
fix: Unwrap long lines in issue templates

fghaas:
Small follow-up to #49:

When creating an issue using the GitHub issue template, and entering preview mode, the issues look odd when they use pre-wrapped lines. To make the experience smoother for issue reporters, unwrap the templates so they use long lines instead.

### DIFF
--- a/.github/ISSUE_TEMPLATE/functional_bug.md
+++ b/.github/ISSUE_TEMPLATE/functional_bug.md
@@ -9,29 +9,21 @@ assignees: ''
 
 # Summary
 
-Replace this with a short summary (one sentence or short paragraph) of
-the issue you are describing.
+Replace this with a short summary (one sentence or short paragraph) of the issue you are describing.
 
 ## Affected paths
 
-List the URL(s) where you found inaccurate information. You can use
-URLs from [the rendered site](https://docs.cleura.cloud) or from [the
-sources on GitHub](https://github.com/citynetwork/docs), either is
-fine.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/functional_bug.md
```diff
@@ -9,55 +9,41 @@ assignees: ''
...
 ## What I did
 
-Explain the steps you undertook that produced results that are
-inconsistent with the documentation, or that led to errors when you
-followed the documentation.
+Explain the steps you undertook that produced results that are inconsistent with the documentation, or that led to errors when you followed the documentation.
 
-It’s OK to remove or mask credentials, but otherwise try to err on the
-side of sharing more rather than less information.
+It’s OK to remove or mask credentials, but otherwise try to err on the side of sharing more rather than less information.
```
colder-is-better, line 22: Not super important, but a simpler/more common syntax would probably be better here. So, instead of "...try to err on the side of sharing...", I would say "...it's preferable to share...".

colder-is-better (commented):
All is well -- but take a look at a pico-nit that just came up :)

.github/ISSUE_TEMPLATE/functional_bug.md
fghaas: Yeah, not now. Please approve as-is and then we can fix that in a later PR.

PR #51, opened 2022-11-01
docs: Update contrib guide with suggestions for screen shots/dumps

fghaas:
Add a note recommending 1080p resolution for GUI screen shots, and a 100-character width limit for CLI screen dumps.

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -4,6 +4,26 @@ You have two options for editing content: directly in your browser
 using GitHub, or using a Git-based workflow from your local work
 environment.
 
+## Notes on specific content additions
+
+**Screen shots:** If you are contributing a change that contains
+screen shots from {{extra.gui}}, they should use a resolution of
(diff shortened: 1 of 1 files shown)


docs/contrib/modifications.md
```diff
@@ -4,6 +4,21 @@ You have two options for editing content: directly in your browser
 using GitHub, or using a Git-based workflow from your local work
 environment.
 
+## Notes on specific content additions
+
+**Screen shots:** If you are contributing a change that contains
+screen shots from {{extra.gui}}, set your screen resolution to 1080p
+(1920×1080 pixels) before you start creating them.
```
colder-is-better, line 11: That is understandable, but I believe it's a bit unreasonable to ask anyone to change their screen resolution (which, we can safely assume, is already optimal for the user's monitor). What we can ask users to do, though, is to _resize_ their screenshots down to 1080p or close to that _before_ uploading.

colder-is-better (commented):
All is well, I have only one comment to make.

colder-is-better (approved):

PR #53, opened 2022-11-02
fix: Remove residual mentions of OpenStack Victoria and Ceph Nautilus

fghaas:
In the Reference section, the version overview was still mentioning
OpenStack Victoria and Ceph Nautilus. Since these versions are no
longer in use anywhere in Cleura Cloud, remove those mentions.


colder-is-better (approved):
LGTM!

PR #54, opened 2022-11-02
feat: Add flavors to reference section

fghaas:
* Explain the naming convention for flavors.
* Explain the available compute tiers, and mention that not all tiers
  are available in all regions.
* Add links to the feature support matrix, in order to explain how
  features map to compute tiers.

### DIFF
--- a/docs/reference/flavors/index.md
+++ b/docs/reference/flavors/index.md
@@ -1,0 +1,58 @@
+# Flavors
+
+Any server instance running in {{extra.brand}} Cloud has a **flavor**,
+which defines the number of virtual CPU cores, the amount of virtual
+RAM, and other performance-related factors.
+
+
+## Naming convention
+
+Flavor names in {{extra.brand}} follow a convention, which can be
+summarized as `X.YcZgb`:
+
+* `X` stands for a lowercase letter identifying the [compute
+  tier](#compute-tiers), with `b` representing the general-purpose
+  tier. It is always followed by a full-stop (`.`).
+* `Y` stands for the number of virtual CPU cores. This number is
+  always followed by the letter `c`.
+* `Z` stands for the allocated amount of virtual RAM, in
+  [gibibytes](https://en.wikipedia.org/wiki/Gigabyte#Base_2_(binary)). This
+  number is always followed by the string `gb`.
+
+For example, the flavor named `b.4c32gb` would be used for a
+general-purpose compute instance with 4 cores and 32 GiB RAM.
+
+
+## Compute tiers
+
+{{extra.brand}} Cloud defines the following compute tiers:
+
+* `b`: General purpose. This is the default compute tier. Instances
+  launched with matching flavors use highly available network-attached
+  storage. This makes them flexible to migrate within the
+  {{extra.brand}} Cloud infrastructure, without interruption.
+  Some
+  [limitations](../../howto/openstack/cinder/encrypted-volumes/#block-device-encryption-caveats)
+  apply to instances with attached [encrypted
+  volumes](../../howto/openstack/cinder/encrypted-volumes/).
+* `s`: High-performance local storage. Instances
+  launched with matching flavors use local, directly-attached
+  storage. This generally provides higher throughput and lower
+  latency for I/O intensive applications, but instances launched with
+  these flavors must configure their own high availability and data
+  replication.
+* `c`: Dedicated CPU. Instances launched with matching flavors are
+  guaranteed to run on compute hardware where CPU cores are allocated
+  to instances on a one-to-one basis and one virtual core maps
+  directly to a physical CPU core.
+* `g`: Virtual GPU. Instances launched with matching flavors have
+  access to a
+  [GPU](https://en.wikipedia.org/wiki/Graphics_processing_unit).
+
+Some tiers are only available in select {{extra.brand}} Cloud
+regions. For details on tier availability, see the [Feature support
+matrix](../features/index.md).
+
+The general-purpose tier is always available to all {{extra.brand}}
+Cloud customers. For access to other tiers, contact our
+[{{extra.support}}](https://{{extra.support_domain}}/servicedesk).
--- a/docs/reference/features/compliant.md
+++ b/docs/reference/features/compliant.md
@@ -12,10 +12,10 @@
 
 ## Virtualization
 
-|               | Sto1HS                | Sto2HS                |
-| ------------- | ----------------      | --------------------- |
-| Dedicated CPU | :material-close:      | :material-close:      |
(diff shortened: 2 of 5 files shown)


docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Naming convention
+
+Flavor names in {{extra.brand}} follow a convention, which can be
+summarized as `X.YcZgb`:
+
+* `X` stands for a lowercase letter identifying the [compute
+  tier](#compute-tiers), with `b` representing the general-purpose
+  tier. It is always followed by a full-stop (`.`).
```
colder-is-better, line 15: Just pointing out here that, in Cleura Cloud Control Panel at least, `X` is an uppercase letter and always equal to `B`.

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+* `Y` stands for the number of virtual CPU cores. This number is
+  always followed by the letter `c`.
```
colder-is-better, line 17: Again, in Cleura Cloud Control Panel that would be a capital `C`.

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+* `Z` stands for the allocated amount of virtual RAM, in
+  [gibibytes](https://en.wikipedia.org/wiki/Gigabyte#Base_2_(binary)). This
+  number is always followed by the string `gb`.
```
colder-is-better, line 20: In Cleura Cloud Control Panel: That's a `GB`, always.

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+For example, the flavor named `b.4c32gb` would be used for a
```
colder-is-better, line 22: ...or `B.4C32GB`, at least in the Cleura Cloud Control Panel.

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+Any server instance running in {{extra.brand}} Cloud has a **flavor**,
+which defines the number of virtual CPU cores, the amount of virtual
+RAM, and other performance-related factors.
+
+
+## Naming convention
+
+Flavor names in {{extra.brand}} follow a convention, which can be
+summarized as `X.YcZgb`:
```
noonedeadpunk, line 11: α has been lost here :)

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+  these flavors must configure their own high availability and data
+  replication. 
+* `c`: Dedicated CPU. Instances launched with matching flavors are
+  guaranteed to run on compute hardware where CPU cores are allocated
+  to instances on a one-to-one basis and one virtual core maps
+  directly to a physical CPU core. 
+* `g`: Virtual GPU. Instances launched with matching flavors have
+  access to a
+  [GPU](https://en.wikipedia.org/wiki/Graphics_processing_unit).
+
```
colder-is-better, line 47: So, from my account, and irrespective of the region, I only see the `B`-tier. Could this be because my account is somehow limited?

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Compute tiers
+
+{{extra.brand}} Cloud defines the following compute tiers:
+
+* `b`: General purpose. This is the default compute tier. Instances
+  launched with matching flavors use highly available network-attached
+  storage. This makes them flexible to migrate within the
+  {{extra.brand}} Cloud infrastructure, without interruption.
```
noonedeadpunk, line 33: unless encrypted block storage device type is selected

docs/reference/flavors/index.md
```diff
@@ -0,0 +1,50 @@
...
+* `s`: High-performance local storage. Instances
+  launched with matching flavors use local, directly-attached
+  storage. This generally provides higher throughput and lower
+  latency for I/O intensive applications, but instances launched with
+  these flavors must configure their own high availability and data
+  replication.
```
noonedeadpunk, line 39: Btw we get the way to live migrate such instances as well and currently re-deploying computes in the regions to support that - fra1 is ready, kna1 in progress, sto2 planned.

fghaas: Being able to live-migrate an instance with block storage migration doesn't change the fact that the application must take care of its own HA and data availability, because if the VM uses local storage only it's not protected against catastrophic loss of a single compute node.

docs/reference/flavors/index.md
fghaas: Not anymore. :)

docs/reference/flavors/index.md
fghaas: No, I deliberately left that out for now because no such public flavors appear to exist, anywhere (yet).

noonedeadpunk: Then why to describe tiers that does not have any public flavors? As high-performance local storage will have also α at the end. So then it's kind of inconsistent what we're documenting.

docs/reference/flavors/index.md
noonedeadpunk: You just mentioned in previous section {{migrate within the Cloud infrastructure, without interruption}} which I read as refference to live migration and not to shared storage. As if we're talking there about failover during hypervisor failure there will be interruptions.
But yes, you're totally right about this section - if there was no intention to talk about ability to live migrate instances it should be left as is.

docs/reference/flavors/index.md
noonedeadpunk: we have B only as public-available tier as of today, correct. Everything else is open only for selected tenants.

docs/reference/flavors/index.md
fghaas: ... unless customers [follow these suggestions](https://github.com/citynetwork/docs/blob/main/docs/howto/openstack/cinder/encrypted-volumes.md#block-device-encryption-caveats).

fghaas: I've added a note to cover this.

docs/reference/flavors/index.md
fghaas: I'll defer adding this until a later date.

fghaas:
I think this is now as complete as it will get, for the time being. @colder-is-better, can I ask you for one more review please?

colder-is-better (approved):

fghaas:
@noonedeadpunk, do you see anything in this PR right now that is functionally *incorrect* (not *incomplete*) and that should keep us from merging?

noonedeadpunk (approved):

PR #58, opened 2022-11-04
feat: Add "A note on commit messages" to contribution guide

fghaas:
Outline our expectations/suggestions for commit messages, and refer to
two helpful articles for additional background information.

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -115,3 +115,31 @@ cd {{ config.extra.brand | lower }}-docs
 export DOCS_ENABLE_HTMLPROOFER=false
 tox -e serve
 ```
+
+## A note on commit messages
+
+When you submit a change, you will need to provide a commit message,
+which is very nearly as important as the change itself. Excellent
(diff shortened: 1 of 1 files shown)


docs/contrib/modifications.md
```diff
@@ -115,3 +115,31 @@ cd {{ config.extra.brand | lower }}-docs
 export DOCS_ENABLE_HTMLPROOFER=false
 tox -e serve
 ```
+
+## A note on commit messages
+
+When submitting a change, you will need to provide a commit message,
```
colder-is-better, line 121: Drop the first comma, maybe?

colder-is-better (commented):
Looking good -- I have a question/suggestion to make only!

docs/contrib/modifications.md
fghaas: Nope, but on second thought phrasing this as "When you submit a change" is better. I've updated the PR.

colder-is-better (approved):

PR #60, opened 2022-11-04
chore: Temporarily disable htmlproofer in GitHub Actions workflows

fghaas:
We are currently getting false positives from the htmlproofer plugin
when it runs from GitHub Actions workflows, in that it reports links
as dead that are evidently alive. This is probably due to bandwidth
throttling or DDOS protection misfiring on the target sites.

This patch should be reverted as soon as the issues subside.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
         run: |
           pip install tox tox-gh-actions
       - env:
+          DOCS_ENABLE_HTMLPROOFER: false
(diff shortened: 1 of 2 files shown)


fghaas:
@colder-is-better I'd appreciate a quick review on this one, since it's currently blocking a couple of other pending patches.

colder-is-better (approved):
Excellent, I had this very issue earlier today and had to manually set `DOCS_ENABLE_HTMLPROOFER` to `false`. So let's make this the default and we see how it goes.

PR #61, opened 2022-11-04
chore: Re-enable htmlproofer in GitHub Actions workflows

fghaas:
We want to use the htmlproofer plugin for deployments, and only
disabled it in GitHub Actions because we were getting false positives
(presumably due to bandwidth throttling or DDOS protection on the
target sites).

This reverts commit 9454deec5081e74531fa37adbc7a23c39758c551.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,6 @@ jobs:
         run: |
           pip install tox tox-gh-actions
       - env:
-          DOCS_ENABLE_HTMLPROOFER: false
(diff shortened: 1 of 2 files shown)


fghaas:
@colder-is-better FYI: This will remain a Draft until the tests succeed again. :)

colder-is-better (approved):
Looking good!

fghaas:
> Looking good!

Yes, seems that way. I've just rebased this on current `main` and I think we can pop this out of the Draft status now.

PR #62, opened 2022-11-07
Reorder pages in Nova howto section, rephrase config-drive howto

fghaas:
- refactor: Reorder pages in OpenStack Nova how-to section
- refactor: In config-drive howto, say "server" rather than "instance"


colder-is-better (approved):
And in an instant, gone are the instances! (Everything LGTM.)

PR #64, opened 2022-11-07
chore: Drop "extra." and "config.extra." prefixes from macros

fghaas:
Prior to this change, we referred to macros defined in the `extra` dictionary of `mkdocs.yml` either with the `extra.` or even the
`config.extra.` prefix.

As [the mkdocs-macros-plugin documentation](https://mkdocs-macros-plugin.readthedocs.io/en/latest/pages/#configuration-variables
) explains, this is actually not necessary: any variables defined in `extra` can be referred to without such a prefix, and it is only when we want to access other variables in `config` (such as `config.repo_url`) that we need to fully qualify them.

Since the documentation sources become much more concise that way, use the abbreviated (unqualified) syntax.

Note: referring to the variables in this manner obviously means that in the `extra` dictionary we cannot have any top-level keys named `config`, `page`, `navigation`, `environment`, `plugin`, `git`, or `files`, but we don't at this time, and avoiding those top-level keys should not be a big issue in the future.

### DIFF
--- a/docs/concepts/index.md
+++ b/docs/concepts/index.md
@@ -1,9 +1,9 @@
 # Concepts and Background
 
 This section contains conceptual and background information about
-{{config.extra.brand}} and the services we support. This is not so
+{{brand}} and the services we support. This is not so
 much about **how** you do something (for that, please see our [How-To
 Guides](../howto/)), more about **why** things work in
 a certain way, or why we have implemented them in a particular way in
-{{config.extra.brand}}.
+{{brand}}.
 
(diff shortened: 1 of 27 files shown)


colder-is-better (approved):
Very nice -- all replacements look good to me!

fghaas:
I've added one more functional improvement, in 6a72848f42825f8d70807318ba16c835d87775e6. Please review that as well, and if you're happy with it please lodge your final approval.

colder-is-better (approved):

PR #68, opened 2022-11-08
fix: Correct information on Dedicated CPU and Virtual GPU availability

fghaas:
In the Sto2 region, a feature that is currently rolling out is
Dedicated CPU, not Virtual GPU. Fix the reference information
accordingly.


colder-is-better (approved):
LGTM!

PR #70, opened 2022-11-14
chore: Fix whitespace in security groups how-to guide

fghaas:
Wrap lines, and replace tabs with spaces.


colder-is-better (approved):
Everything looks good to me.

PR #71, opened 2022-11-17
Update links to CDPA and GTC documents

Skipper126:
CDPA and GTC documents were migrated from citycontrolpanel.com to object storage. The "Create Account" page had to be updated with new links.


fghaas (commented):
In principle this is fine; the commit message still needs line-wrapping in order to pass the `gitlint` check.

However, I wonder why these documents still say City Network in both the file names and the content? Shouldn't they be replaced with versions that say Cleura?

Skipper126:
Cleura GTC and CDPA documents are not created yet, that's why we still use CityNetwork name there. But main reason for this PR is that we had to move the documents to another place, since we plan to shut down citycontrolpanel.com, under which they were previously stored.

fghaas (approved):

PR #72, opened 2022-11-21
Adjust "Creating security groups" page after functionality changes

Skipper126:
The how-to page for "Creating security groups" had section about removing "default" security group ingress rules, that allowed all traffic. This was recently changed in CCMP, and for new accounts the default ingress rules are only allowing traffic from the same security group.

Now, the "remove security group rules" section does not recommend removing the rules, but only suggests that it is something that you can do, if you for any reason want to.

Screenshots and CLI output was adjusted as well.

The screenshot showing remaining rules, after removal of default ingress rules, had wrong set of rules removed. This was changed, and the screenshot filename was fixed.


fghaas (changes requested):
Hmmm, I don't think we can land this as-is. We only recently changed the defaults for the ingress rules in the `default` group. That means that we still have a bunch of `default` groups out there with the ingress rules being permissive. If we now document that the ingress rules are closed-by-default, those users will check this documentation and assume that their `default` rules are restrictive when in fact they are still permissive, on account of their `default` group being older than the server-side default change.

You will need to cover both cases: a restrictive-by-default `default` group, and a permissive-by-default `default` group.

Please update your topic branch accordingly.

fghaas:
@Skipper126 just checking, will you return to this before the end of the week?

Skipper126:
Sorry for leaving this hanging. Yes, I'll try to finish it this week.

To follow up on the discussion.

Instead of having separate sections for each case, what do you think about having a warning box, saying something like "If your account has been created before X date, your "default" security group ingress rules are allowing all traffic by default. See the "remove security group rules" section to learn how to configure it properly."

Then, the "remove security group rules" section would still show the "new" examples, but could mention the "old" case: "For accounts created before X date, the ingress rules allow all traffic".

This way we could keep the docs following the "new" way, while still making it clear, that there are some "legacy" settings around, and who is affected. And with time, this would be less and less relevant, and probably in a year we could remove the warnings.

fghaas:
> Instead of having separate sections for each case, what do you think about having a warning box, saying something like "If your account has been created before X date, your "default" security group ingress rules are allowing all traffic by default. See the "remove security group rules" section to learn how to configure it properly."

If you want to take that approach, then you also have to explain how people can check when their security group was created.

fghaas:
> I believe the account creation date is enough here

Why would you instruct a user to look up their account creation date when they can do `openstack security group show default` and check the `created_at` field right there?

Skipper126:
Let's say I write following warning "For accounts created after X date, the default security group accepts all ingress traffic by default. To be sure, check your default security group ingress rules by doing following..."

The source of truth are still the ingress rules. It could be the default group creation date as well, but the rules are more direct.
The account creation date is not a definitive source of truth, but rather an easy to see filter for a reader. If the reader is sure that their account was created after the date, then they won't need to do the checks, and can proceed with documentation main flow. With time, in a few months and onward, this will be more and more common case.
If this is an older account, or created somewhere around the date, or reader is not sure, they can always check the rules as instructed.

My intend is that we need to explain on what rules some users could have each case, and that in a few months, new users won't have to worry about that.

Also, I can understand that I might be overcomplicating this. If you feel that a separate section is simpler and more clear, then I agree on that as well.

fghaas:
Okay, whichever approach you choose, it'll just have to work for both "old" and "new" `default` security groups.

Also, please rebase on current `main` and fix your commit message.

fghaas:
Please don't "rebase" by merging `main` into your topic. Actually rebase your topic branch on `main` instead.

Skipper126:
It appears to be default behaviour for Github "sync fork" button, which I clicked. I'll rebase and squash the commits later on of course.

fghaas:
> It appears to be default behaviour for Github "sync fork" button, which I clicked.

Yeah, the GitHub web UI is unfortunately a bit weird in more ways than one. :)

docs/howto/openstack/neutron/create-security-groups.md
```diff
@@ -309,3 +316,112 @@ All the rules for a simple web server are now in place.
 
 For any additional protocol or ingress rule, simply follow the same
 procedure as above.
+
+## Adjust permissive default security group
+
+If your account was created before 16th November 2022, and you didn't
```
fghaas, line 322: Use ISO8601 dates, please. Thus, put `2022-11-16`, not `16th November 2022`.

docs/howto/openstack/neutron/create-security-groups.md
```diff
@@ -66,13 +66,22 @@ first](/howto/getting-started/enable-openstack-cli).
 ## Removing default ingress rules
 
 By default, a security group named `default` has been already created
-for you, allowing all traffic from any source (ingress), and to any
-destination (egress).
+for you, blocking all traffic from any source (ingress), except from
+servers and ports being in the same security group. All traffic to any
+destination (egress) is allowed by default.
+
+> For accounts created before 16 November 2022, the default security
```
fghaas, line 73: Here too, please use an ISO 8601 date.

fghaas (commented):
I think this is almost good to go now. I just have one more small fix to ask for, in two places.

@colder-is-better, any other thoughts from you, or do you think this is (nearly) good to merge now, too?

colder-is-better (commented):
Looks good to me --- I only have a few femto-nits.

fghaas:
@Skipper126 Just checking: will you be pushing a final update today, or should @colder-is-better take care of the last suggested fixes for you? Either way is fine, just let us know. :)

Skipper126:
> @Skipper126 Just checking: will you be pushing a final update today, or should @colder-is-better take care of the last suggested fixes for you? Either way is fine, just let us know. :)

Hey, I'm on it and will push the fixes within 1h :)

fghaas (approved):
This looks good to me now. I'll go ahead and merge.

PR #57, opened 2022-11-04
Adding old documentation from kb.citynetwork.eu.

Tomas-Rinblad-at-cleura:
I have taken the old documentation from kb.citynetwork.eu regarding how to move instances and rewritten it to the documentation page.

Resource:
https://kb.citynetwork.eu/kb/how-to-articles/openstack/move-a-instance-between-different-places

PR #73, opened 2022-11-21
feat: Add HowTo moving servers between regions

Jeeaaasus:
This is a continuation from PR #57 as I have taken over this guide.
Resource: https://kb.citynetwork.eu/kb/how-to-articles/openstack/move-a-instance-between-different-places


docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+
+    Printout reference:
+
+    ```plain
+    +-----------------+-----------------+---------+-----------------+--------------------+-------------+
+    | ID              | Name            | Status  | Networks        | Image              | Flavor      |
+    +-----------------+-----------------+---------+-----------------+--------------------+-------------+
+    | a51e3d79-....-. | Name2           | SHUTOFF | STO2_test_netwo |                    | b.1c1gb     |
+    | ...-....-...... |                 |         | rk=10.x.y.z     |                    |             |
+    | ......          |                 |         |                 |                    |             |
```
fghaas, line 46: Please don't replace UUIDs with dots. To mask real UUIDs with non-existent ones, just use `uuidgen`.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+    +------------------+-------------------------------------------+
+    | Field            | Value                                     |
+    +------------------+-------------------------------------------+
+    | volumes_attached | id='14942b23-....-....-....-............' |
+    |                  | id='34e2ca96-....-....-....-............' |
+    |                  | id='68756558-....-....-....-............' |
+    +------------------+-------------------------------------------+
+    ```
+
+    If there are multiple volumes attached, the first volume in the list is the server system volume. Copy this ID.
```
fghaas, line 71: It'll be the server system volume *only if* it's a BFV server. So, you'll need to include a check that the server is in fact using a BFV flavor.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Creating a copy of a volume
+
+To make an image out of a server, first you have to make a copy of the system volume, by using the following command:
+
+```bash
+openstack volume create --source <volume_id> <new_volume_name>
```
fghaas, line 78: Okay so `openstack volume create --source <volume_id>` works even if the source volume is `in-use`, whereas `openstack image create --volume <volume_id>` fails unless the volume is `available`?

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Creating an image of a volume
+
+Then create the image, by using the following command:
+
+```bash
+openstack image create --volume <volume_id> <new_image_name>
+```
+
+> Substitute `<volume_id>` with the ID from the newly created volume in the previous step.
```
fghaas, line 117: Don't reuse placeholder names. Earlier, you used `<volume_id>` to refer to the UUID of the server's volume; now you're using it to refer to the UUID of a copy. Use something like `<source_volume_id>` and `<copy_volume_id>` instead.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+
+```plain
++--------+--------+
+| Field  | Value  |
++--------+--------+
+| status | active |
++--------+--------+
+```
+
+> It is now safe to remove the volume you created earlier.
```
fghaas, line 156: Add a command for this. Again, make sure you use the placeholder `<copy_volume_id>`, not `<volume_id>`.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+## Downloading an image
+
+Download the image to the local machine, using the following command:
+
+```bash
+openstack image save --file <local_image_name>.<disk_format> <image_id>
+```
+
+> Substitute `<disk_format>` with the value from the printout in the previous step.  
+> Substitute `<image_id>` with the ID from the printout in the previous step.
```
fghaas, line 167: The previous step is currently `openstack image show -c status <image_id>` which doesn't include the disk format, nor the image UUID, in its printout.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,221 @@
...
+
+When you have downloaded the file, you are done with the steps for the source region. The following steps will be done on the target region.
+
+> It is now safe to remove the image you created earlier.
```
fghaas, line 171: Yeah, maybe it's not. :) You don't know whether the image you downloaded is broken or not. Maybe try `md5sum` against the downloaded file, and compare it to the image's `checksum` value?

fghaas (commented):
I have a few comments; please find them inline.

docs/howto/openstack/nova/move-server-between-regions.md
Jeeaaasus: Correct.

And `openstack image create --force --volume <volume_id> <new_image_name>` gives this really odd error message:
```
'--force' was given, which is not supported when creating an image from a volume. This will be an error in a future version.
```
Even though it says in the CLI help to use that with volumes:
```
--force       Force image creation if volume is in use (only meaningful with --volume)
```

It's odd..

fghaas: Very odd indeed. But it is what it is. :)

Jeeaaasus:
Adding this for myself to not forget to:
add optional step to shutoff the server before making the copy of the system volume.
add the steps necessary to also transfer all the volumes the server has attached.

docs/howto/openstack/nova/move-server-between-regions.md
Jeeaaasus: I cannot find any command to have openstack display the checksum of an image.

`openstack image show -c checksum <image_id>`
```
No recognized column names in ['checksum']. Recognized columns are ('container_format', 'created_at', 
'disk_format', 'file', 'id', 'min_disk', 'min_ram', 'name', 'owner', 'properties', 'protected', 'schema', 
'size', 'status', 'tags', 'updated_at', 'visibility').
```

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,311 @@
+# Moving a server from one region to another
+
+This guide will show you how to move a server to a different region in {{brand}} Cloud.
```
fghaas, line 3: With #81 having landed, please change `{{brand}} Cloud` to just `{{brand}}` when you next rebase on current `main`.

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,311 @@
...
+| protected        | False                                                                         |
+| schema           | /v2/schemas/image                                                             |
+| status           | saving                                                                        |
+| tags             |                                                                               |
+| updated_at       | 2022-11-17T16:26:58Z                                                          |
+| visibility       | shared                                                                        |
++------------------+-------------------------------------------------------------------------------+
+```
+
+But your image is not yet ready to use, {{brand}} Cloud still needs to process the file, which shouldn't take long. To check the status of the image use the ID of the new image with the following command:
```
fghaas, line 265: Here too, please replace `{{brand}} Cloud` with just `{{brand}}`.

fghaas (commented):
Just a small note here, for when you next rebase and update this PR.

fghaas:
No I think that had not so much to do with GitHub but rather with www.conventionalcommits.org temporarily glitching. :)

docs/howto/openstack/nova/move-server-between-regions.md
```diff
@@ -0,0 +1,364 @@
...
+This guide will show you how to move a server to a different region in {{brand}}.
+
+## Prerequisites
+
+In order to move a server from one region to another, you will need
+
+- a correctly installed and configured [OpenStack CLI client](../../getting-started/enable-openstack-cli.md),
+- access to the RC files with credentials for both the source and target region,
+- enough space on the local machine to be able to download an image of your server,
+- a sufficiently configured target region including a [virtual network](../neutron/new-network.md) and necessary [security groups](../neutron/create-security-groups.md).
```
fghaas, line 12: I have pushed a suggested reformat/rewrite for this bullet list in [d9790d6](https://github.com/citynetwork/docs/pull/73/commits/d9790d62cefa9fe9b017e70386d100a9495b9fd6), which we can subsequently squash into the top commit in this PR.

fghaas (commented):
I think this now mostly good to go — we might want to move the bit on creating an image from a volume to the Glance howto section, and the one on creating a volume from an image to the Cinder one, but we can do that  in a later refactoring.

Unless there are any objections, I will go ahead and squash the two fixup commits into the main commit, and then I'll go ahead and merge this.

Jeeaaasus:
Sounds great, no objections from me!

fghaas:
Hmmm this is very odd. I have no idea why the checks aren't reporting back (they did all complete successfully).

fghaas:
@colder-is-better, as far as I am concerned this PR is ready to merge now. Any objections?

colder-is-better:
We're good to go, I believe :)

fghaas (approved):

PR #74, opened 2022-11-23
feat: Add HowTo resizing servers

Jeeaaasus:
This guide goes through the basic steps to resize a server, this includes, listing flavors, performing a resize, confirming a resize. Both through the GUI and OS CLI.


docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+    +-----------------------+---------------+
+    ```
+
+The resize process might take a minute or more. {{brand}} Cloud will now make a restore point incase the resize process fails, it would then restore your server to the state it was before the resize.
+
+## Confirming the resize
+
+Your server is now using the new flavor you selected earlier and you need to make sure the server is working as intended after the resize.
+
+Once you are certain you server is working as intended, you **must** confirm the resize. If you fail to confirm the resize, your server will have the restore point created by the resize process left behind and this can cause your server to be reverted to this point at a later date, resulting in **data loss**.
```
noonedeadpunk, line 105: Well, you actually should - not must. As resize will be auto-confirmed in case it has not be reverted in 24hours.
Also this side effect for reverting is valid only for non-bfv which is not the way we expect new servers to be created and usage of non-bfv should reduced as much as possible anyway. Moreover, we provide instruction that explicitly fetches only bfv flavors.

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
+# Resizing a server
+
+This guide will walk you through the required steps to change the number of CPU cores and the amount of memory your server has access to, this is done my changing the server's [flavor](/reference/flavors/).
+
+[Resize](https://docs.openstack.org/nova/latest/admin/configuration/resize.html) (or Server resize) is the ability to change the flavor of a server, thus allowing it to upscale or downscale according to user needs. A resize operation is a two-step process for the user: the first step is to perform the resize, and the second step is to either confirm (verify) success and release the old server or to declare a revert to release the new server and restart the old one.
+
+## Prerequisites
+
+You need to have a server you wish to resize. Additionally, if you prefer to work with the OpenStack CLI, then make sure to properly [enable it first](/howto/getting-started/enable-openstack-cli/).
```
noonedeadpunk, line 9: I find enabling CLI as super weird statement. Can we re-phrase it to create CLI user or smth as we don't disable CLI/API - we just don't provide access details on account creation.

docs/howto/openstack/nova/resize-server.md
Jeeaaasus: oh really, there's no roll back when reverting while using BFV. I did not know that, that's great. Will rewrite this whole section.

docs/howto/openstack/nova/resize-server.md
Jeeaaasus: I agree, however, this is the exact same language as used in many other guides and also in the referenced "getting started" guide.

fghaas: "Enabling" means "making something work". In order to make the `openstack` CLI work, you need a user, a password, and properly configured environment variables. Without the foregoing, it is useless. So, the wording is fine.

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Listing available flavors
+
+=== "{{gui}}"
+    Navigate to the server list.
+
+    ![navigation-panel](assets/resize-server/01-left-side-panel.png)
```
fghaas, line 16: Please add a proper [alt text](https://accessibility.huit.harvard.edu/describe-content-images) here (this comment applies to all inline images in this PR).

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+
+    ```plain
+    X.1c1gb
+    X.1c2gb
+    X.1c4gb
+    X.2c2gb
+    X.2c4gb
+    X.2c8gb
+    X.2c16gb
+    X.4c4gb
```
fghaas, line 46: Rather than the `X` prefix, you might as well use `B` which constitutes the vast majority of our current flavors.

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+    To start the resize use the following command:
+
+    ```bash
+    openstack server resize --flavor <new_flavor> <server_id>
+    ```
+
+    While the resize is ongoing the server should have the OS-EXT-STS:task_state of `resize_migrating` and the status of `RESIZE`.
+
+    ```bash
+    openstack server show -c OS-EXT-STS:task_state -c OS-EXT-STS:vm_state -c status <server_id>
```
fghaas, line 74: I may be wrong here, but I think non-admin accounts don't get to see the `OS-EXT-STS:foo` attributes. Please verify.

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
+# Resizing a server
+
+This guide will walk you through the required steps to change the number of CPU cores and the amount of memory your server has access to, this is done my changing the server's [flavor](/reference/flavors/).
+
+[Resize](https://docs.openstack.org/nova/latest/admin/configuration/resize.html) (or Server resize) is the ability to change the flavor of a server, thus allowing it to upscale or downscale according to user needs. A resize operation is a two-step process for the user: the first step is to perform the resize, and the second step is to either confirm (verify) success and release the old server or to declare a revert to release the new server and restart the old one.
```
fghaas, line 5:
```markdown
A resize operation is a two-step process for the user:
1. initiate the resize,
2. either confirm (verify) success and release the old server, or
   declare a revert to release the new server and restart the old one.
```

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+    | OS-EXT-STS:vm_state   | resized       |
+    | status                | VERIFY_RESIZE |
+    +-----------------------+---------------+
+    ```
+
+The resize process might take a minute or more. {{brand}} Cloud will now make a restore point incase the resize process fails, it would then restore your server to the state it was before the resize.
+
+## Confirming the resize
+
+Your server is now using the new flavor you selected earlier and you need to make sure the server is working as intended after the resize.
```
fghaas, line 103: Insert comma before "and".

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,125 @@
...
+    +-----------------------+------------------+
+    | Field                 | Value            |
+    +-----------------------+------------------+
+    | OS-EXT-STS:task_state | resize_migrating |
+    | OS-EXT-STS:vm_state   | active           |
+    | status                | RESIZE           |
+    +-----------------------+------------------+
+    ```
+
+    You may proceed with the next step once your server status is `VERIFY_RESIZE`
```
fghaas, line 87: Missing punctuation. Add a full-stop or a colon at the end.

fghaas (commented):
Overall this looks good! I have a few comments; please take a look.

docs/howto/openstack/nova/resize-server.md
Jeeaaasus: nono, I went through these steps on my non-admin account to get these printouts.

docs/howto/openstack/nova/resize-server.md
Jeeaaasus: I thought it would be best to use the same naming convention as the official flavor reference document, no? https://docs.cleura.cloud/reference/flavors/#naming-convention

I could skip the fake printout and simply describe that the list will be grouped by compute type and sorted by cores and then ram and printed in the X.YcZgb form. If you prefer that instead.

docs/howto/openstack/nova/resize-server.md
Jeeaaasus: I'm not certain what you are asking me do, am I totally wrong with this example?
```
    ![The left hand side navigation panel with the word, servers, highlighted.](assets/resize-server/01-left-side-panel.png)
```

fghaas: Not totally wrong at all. :) As a minor improvement I'd put:

```markdown
![The left hand side navigation panel, with the word "Servers" highlighted.](assets/resize-server/01-left-side-panel.png)

File: docs/howto/openstack/nova/resize-server.md (hunk 4 above)
Comment (fghaas) [reply docs/howto/openstack/nova/resize-server.md:46]: The reference is a reference. This how-to is a how-to. :)

I think the printout is fine, but it shouldn't use a prefix that doesn't actually exist in any flavor. Hence the suggestion to use `b.` and not `X.`.

File: docs/howto/openstack/nova/resize-server.md (hunk 6 above)
Comment (fghaas) [reply docs/howto/openstack/nova/resize-server.md:5]: Please mark conversations as "resolved" only under the following circumstances:

1. The issue described is fixed in the topic branch (not just locally in your checkout).
2. The issue doesn't _need_ fixing.

I don't see either of these conditions applying here, so I've unresolved the conversation.

Comment (Jeeaaasus) [reply docs/howto/openstack/nova/resize-server.md:5]: I understand.

Comment (Jeeaaasus) [reply docs/howto/openstack/nova/resize-server.md]: Is the proper etiquette that the reviewer marks as resolved?

Comment (fghaas) [reply docs/howto/openstack/nova/resize-server.md]: Not necessarily; it's perfectly okay for you to mark it as resolved — after you've pushed the fix. :)

File: docs/howto/openstack/nova/resize-server.md [hunk 9]
```diff
@@ -0,0 +1,131 @@
+# Resizing a server
+
+This guide will walk you through the required steps to change the number of CPU cores and the amount of memory your server has access to, this is done my changing the server's [flavor](/reference/flavors/).
```
fghaas, line 3: Typo, `s/my/by/`.

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,131 @@
...
+You need to have a server you wish to resize. Additionally, if you prefer to work with the OpenStack CLI, then make sure to properly [enable it first](/howto/getting-started/enable-openstack-cli/).
+
+## Listing available flavors
+
+=== "{{gui}}"
+    Navigate to the server list.
+
+    ![The left hand side navigation panel, with the word "Servers" highlighted.](assets/resize-server/01-left-side-panel.png)
+
+    Find the server you want to resize in the list and on the right hand side click on it's menu button.
```
fghaas, line 22: * Insert comma before "and".
* Use "right-hand side".
* Replace "it's" with "its".

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,131 @@
...
+    
+    ![An orange circle with three white dots.](assets/resize-server/02-menu-button.png)
+
+    Click on _Modify Server_.
+
+    ![A list of server actions, with the line "Modify Server" highlighted.](assets/resize-server/03-menu-list.png)
+
+    On this panel near the top, find the section called _Flavor_, this is the current flavor used by the server. Press on the dropdown menu to see all available flavors.
```
fghaas, line 30: Split the first sentence into two.

```markdown
On this panel near the top, find the section called _Flavor_. This is the current flavor used by the server.
```

docs/howto/openstack/nova/resize-server.md
```diff
@@ -0,0 +1,131 @@
...
+    +-----------------------+---------------+
+    | Field                 | Value         |
+    +-----------------------+---------------+
+    | OS-EXT-STS:task_state | None          |
+    | OS-EXT-STS:vm_state   | resized       |
+    | status                | VERIFY_RESIZE |
+    +-----------------------+---------------+
+    ```
+
+The resize process might take a minute or more. {{brand}} Cloud will now make a restore point incase the resize process fails, it would then restore your server to the state it was before the resize.
```
fghaas, line 105: It's "in case", not "incase". Also, split the second sentence into two:

```markdown
{{brand}} Cloud will now make a restore point in case the resize process fails.
It would then restore your server to the state it was before the resize.
```

fghaas (commented):
This looks almost good to merge now. Just a few remaining small nits.

Jeeaaasus:
Changes pushed, anything other thoughts now? :)

fghaas:
> Changes pushed, anything other thoughts now? :)

Just wrapping lines and fixing whitespace, and an update to the commit message. I've pushed that in a separate fixup commit ([72021e2](https://github.com/citynetwork/docs/pull/74/commits/72021e2986662741b8bca4d02fab509b3a305f0c)), which I intend to squash into yours unless you have objections.

Jeeaaasus:
No.

Jeeaaasus:
Thank you for the help as always!

PR #76, opened 2022-11-27
feat: Add HowTo: Site-to-Site VPN w/ Neutron VPNaaS

colder-is-better:
We show, step by step, how to establish a Site-to-Site IPSec VPN connection between two regions, using the Cleura Cloud Management Cloud or the OpenStack CLI.


docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+	```bash
+	apt install python3-neutron-vpnaas
+	```
+=== "Mac OS X with Homebrew"
+    This Python module is unavailable via `brew`, but you can 
+    install it via `pip`.
+=== "Python Package"
+	```bash
+	pip install neutron-vpnaas
+	```
```
fghaas, line 30: Errr, no. `neutron-vpnaas` is the *server* component implementing the VPNaaS API; it is not needed for client-side API VPNaaS management.

The VPNaaS client functionality is in `python-neutronclient`, which is a hard dependency of `python-openstackclient`. So as long as the OpenStack client is installed and the server supports Neutron VPNaaS, no additional package installation on the client is required.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+    ```
+    
+    ### Instantiate a pre-shared key
+       
+    Before establishing a site-to-site IPSec VPN connection, you must 
+    have a randomly generated pre-shared key. You may use `openssl` for 
+    generating a random string and immediately set it to a shell variable:
+        
+    ```bash
+    PRE_SHARED_KEY=$(openssl rand -hex 24)
```
fghaas, line 368: Does this _have_ to be a 24-byte hex key? If it does not, then please explain that this is an example and that you could also use, say, `pwgen 64 1` for the same purpose.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+    openstack vpn ipsec site connection list
+    ```
+        
+    ```plain
+    +--------------------------+------------------+---------------+--------------------------+--------+
+    | ID                       | Name             | Peer Address  | Authentication Algorithm | Status |
+    +--------------------------+------------------+---------------+--------------------------+--------+
+    | 5f44be31-0588-4f33-883d- | vpn-conn-to-kna1 | 203.0.113.101 | psk                      | ACTIVE |
+    | 9e2d97be55e1             |                  |               |                          |        |
+    +--------------------------+------------------+---------------+--------------------------+--------+
```
fghaas, line 498: Looks like this missing a `CLIFF_MAX_WIDTH=100` setting; is this intentional?

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+
+## Testing the site-to-site VPN connection
+
+One way to test the VPN connection is to have two servers (e.g., 
+`server-fra1` and `server-kna1`), each on a different region (e.g., 
+`fra1` and `kna1` respectively), ping each other using private IP 
+addresses. With no VPN connection between the two regions, pinging 
+should not be possible:
+
+```bash
```
fghaas, line 545: This should use `console` syntax highlighting, not `bash`.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+ubuntu@server-fra1:~$ ping -c 3 10.15.20.148
+PING 10.15.20.148 (10.15.20.148) 56(84) bytes of data.
+
+--- 10.15.20.148 ping statistics ---
+3 packets transmitted, 0 received, 100% packet loss, time 2050ms
+```
+
+```bash
```
fghaas, line 553: `s/bash/console/`

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+PING 10.15.25.58 (10.15.25.58) 56(84) bytes of data.
+
+--- 10.15.25.58 ping statistics ---
+3 packets transmitted, 0 received, 100% packet loss, time 2045ms
+```
+
+On the other hand, with a VPN connection established between the two 
+regions, pinging should be all possible:
+
+```bash
```
fghaas, line 564: `s/bash/console/`

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,586 @@
...
+64 bytes from 10.15.20.148: icmp_seq=1 ttl=62 time=32.8 ms
+64 bytes from 10.15.20.148: icmp_seq=2 ttl=62 time=32.5 ms
+64 bytes from 10.15.20.148: icmp_seq=3 ttl=62 time=32.6 ms
+
+--- 10.15.20.148 ping statistics ---
+3 packets transmitted, 3 received, 0% packet loss, time 2003ms
+rtt min/avg/max/mdev = 32.544/32.624/32.761/0.096 ms
+```
+
+```bash
```
fghaas, line 576: `s/bash/console/`

fghaas (changes requested):
Overall I think this is excellent!

I have a few minor suggestions, and one larger one: should we perhaps add some information on how to temporarily tear down the VPN connection, and how to re-instate it thereafter?

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: I mixed things up a bit, and now I see that I should not have installed `python3-neutron-vpnaas` (Debian/Ubuntu) or `neutron-vpnaas` (Mac OS). But since I clearly remember the command `vpn` being unavailable for `openstack`, I took a closer look and realized that `python-neutronclient` is *not* a hard dependency of `python-openstackclient`. Actually, in Debian or Ubuntu, in addition to `python3-openstackclient`, you have to install `python3-neutronclient`. And in `Mac OS`, in addition to `python-openstackclient`, you have to install `python-neutronclient`. I will update the HOWTO accordingly.

fghaas: Oh, interesting! Maybe that dependency changed (semi-)recently and I missed it. At any rate, the Neutron _client_ library it is. :)

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: No, no, it does not have to be a 24-byte hex key -- that's just an example I came up with, God-knows-why. I will clarify that and also mention (the fine) `pwgen`, which I use and love anyways.

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: Um, no, `CLIFF_MAX_WIDTH=100` is actually in both RC files I use.

fghaas: Odd. Perhaps Cliff just doesn't want to ever wrap header cells, and thus the `Authentication Algorithm` cell pushed the line width over the limit.

colder-is-better:
Indeed, temporarily shutting down the connection and re-instating it when needed would be really useful. I will look into it and update the HOWTO accordingly.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,727 @@
...
+    +--------+--------+
+    ```
+        
+    You get the same status by issuing a similar command from the right 
+    side.
+        
+    > Again, since you may have to wait several seconds before seeing 
+    the status change you expect, try something like this: `watch -n 3 
+    "openstack vpn ipsec site connection show vpn-conn-to-kna1 -c Status"`. 
+    Hit CTRL+C to stop watching.
```
fghaas, line 727: Here's how you get this properly formatted:

```markdown
    > Again, since you may have to wait several seconds before seeing 
    > the status change you expect, try something like this:
    > ```bash
    > watch -n 3 "openstack vpn ipsec site connection show vpn-conn-to-kna1 -c Status"
    > ```
    > Hit CTRL+C to stop watching.
```

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,727 @@
...
+    | Route Mode               | static                                             |
+    | State                    | True                                               |
+    | Status                   | ACTIVE                                             |
+    | VPN Service              | d74d51f0-182d-4d88-952a-1d593ce696fd               |
+    | dpd                      | {'action': 'hold', 'interval': 30, 'timeout': 120} |
+    | project_id               | dfc700467396428bacba4376e72cc3e9                   |
+    +--------------------------+----------------------------------------------------+
+    ```
+
+## Testing the site-to-site VPN connection
```
fghaas, line 545: Now that there is a section on enabling and disabling a connection, I would move this entire section down to the very end (because it also serves as a test of whether the connection has been successfully disabled and re-enabled).

fghaas (commented):
Looks good; two more minor suggestions. When fixed, we can merge this.

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: Updated, thanks :)

fghaas: You can probably remove `-n 3` as well, because it doesn't make that big of a difference if we refresh every 2 seconds (default) or every 3. :)

colder-is-better:
Well, at first I thought of that too, but then I changed my mind. My rationale is this; I imagine someone reading through this rather long HOWTO, feverishly typing away command after command until they have their fine site-to-site VPN connection established. If that someone was anything like me, then I bet they would want to test if the connection works as advertised sooner rather than later. That's why I believe it makes sense to leave this section there. What do you think?

fghaas:
Fair point; in that case however I would suggest you reprise the `ping` command examples one more time each (outside the content tabs) in the sections on disabling and re-enabling the connection. Just to remind people how they can test that the connection is effectively down, and back up.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,741 @@
...
+```
+
+## Disabling an active connection
+
+You may, at any time, disable an active site-to-site VPN connection.
+As soon as you disable such a connection, any two servers residing in
+different regions of the previously established connection will be
+unable to reach each other by private IP addresses. You will be able
+to verify that, e.g., by pinging, as we demonstrate in the section
+regarding testing a site-to-site connection.
```
fghaas, line 603: **No.** Don't make readers jump around. Just restate the `ping` command, *at the end* of this section. The connection is down per the API, you ping to verify that it is in fact down. Simple and logical.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,741 @@
...
+        
+    ```plain
+    +--------+-------+
+    | Field  | Value |
+    +--------+-------+
+    | Status | DOWN  |
+    +--------+-------+
+    ```
+    
+    It turns out that this direction of the connection is also disabled.
```
fghaas, line 676: After this (outside the content tab) is where you put the instructions for verifying with `ping`.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,741 @@
...
+
+## Enabling an inactive connection
+
+You can easily enable an inactive site-to-site VPN connection. As soon
+as you enable such a connection, any two servers residing in different
+regions of the established connection will be again able to reach each
+other by private IP addresses. You will be able to verify that, e.g.,
+by pinging, as we demonstrate in the section regarding testing a
```
fghaas, line 684: Same consideration as above.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,741 @@
...
+        
+    You get the same status by issuing a similar command from the right 
+    side.
+        
+    > Again, since you may have to wait several seconds before seeing 
+    the status change you expect, try something like this: 
+    ```bash
+    watch "openstack vpn ipsec site connection show vpn-conn-to-kna1 -c Status"
+    ``` 
+    Hit CTRL+C to stop watching.
```
fghaas, line 741: Again, here (outside the content tab) is where you restate the `ping` command, this time to verify that the connection is back.

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,786 @@
...
+3 packets transmitted, 0 received, 100% packet loss, time 2050ms
+```
+
+## Enabling an inactive connection
+
+You can easily enable an inactive site-to-site VPN connection.
+
+=== "{{gui}}"
+    Currently, there is no way to enable an inactive connection from 
+    the {{gui}}. If you want to disable an inactive connection, please use
```
fghaas, line 702: Apologies, just caught this now in final review: `s/disable/enable/` (or maybe use "re-enable").

docs/howto/openstack/neutron/vpnaas.md
```diff
@@ -0,0 +1,786 @@
...
+    watch "openstack vpn ipsec site connection show vpn-conn-to-kna1 -c Status"
+    ``` 
+    Hit CTRL+C to stop watching.
+
+To test that a previously disabled site-to-site connection is now enabled,
+select a server from one region and try to ping a server in another region.
+Suppose, for example, that a re-enabled connection involves regions
+`fra1` and `kna1`, and that we have servers `server-fra1` (in `fra1`) and
+`server-kna1` (in `kna1`). With the VPN connection between the two regions
+now re-established, pinging should be all possible:
```
fghaas, line 761: I think the word "all" can be removed here.

fghaas (commented):
Looks good now! Two more trivial fixes, and please update the commit message subject to say "with" rather than just "w/".

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: Aha, good catch! I say we go with "re-enable".

docs/howto/openstack/neutron/vpnaas.md
colder-is-better: It sure can :)

fghaas (approved):
Excellent. I'll go ahead and merge this now.

PR #77, opened 2022-11-29
refactor: Slightly reorganize Kubernetes how-tos

fghaas:
* Add `index.md` files in the `howto/kubernetes/magnum` and `howto/kubernetes/gardener` directories.
* Move generic information about Gardener and Magnum into those files.
* Clarify that Gardener is in closed beta, and add instructions on how to enroll in it.

### DIFF
--- a/docs/howto/kubernetes/index.md
+++ b/docs/howto/kubernetes/index.md
@@ -1,40 +1,9 @@
 # Kubernetes in Cleura
 
-<!-- Config interpolation for the company name didn't work for the 
-title here. -->
+In {{brand}} Cloud you have several options for deploying and managing
+Kubernetes clusters.
 
-In {{company}} you can run Kubernetes in various
-ways. [{{gui}}](https://{{gui_domain}}) includes management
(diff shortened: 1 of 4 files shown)


docs/howto/kubernetes/index.md
```diff
@@ -1,40 +1,9 @@
 # Kubernetes in Cleura
 
-<!-- Config interpolation for the company name didn't work for the 
-title here. -->
+In {{brand}} Cloud you can run Kubernetes in several
+ways.
```
colder-is-better, line 5: Since you don't _run_ Kubernetes, I would rephrase this sentence as follows: 

In {{brand}} Cloud there are several ways to deploy a Kubernetes cluster.

colder-is-better (commented):
Looks good --- I have just one suggestion.

docs/howto/kubernetes/index.md
fghaas: Fair point; I used not quite that but something similar. :)

PR #79, opened 2022-11-30
build: Bump version tags for GitHub actions

fghaas:
GitHub advises against the use of actions using Node.js 12. This is true for the v2 releases of the `checkout` and `setup-python`
actions. Thus, bump those actions to their latest versions (v3 and v4, respectively).

While we're at it, bump the `upload-artifact` action to v3 as well.

Reference:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


colder-is-better (commented):
All version bumps seem consistent to me :)

fghaas:
OK @colder-is-better, I think this is good for approval now.

colder-is-better (approved):

PR #80, opened 2022-12-05
feat: Add how-to guides for working with the Swift API

fghaas:
Add how-to guides for

* creating a private Swift container,
* creating a public Swift container,
* defining object expiry,
* creating TempURLs,
* using versioned containers.

In addition, add an index page and pages.md for the S3 API section.


docs/howto/object-storage/swift/private-container.md
```diff
@@ -0,0 +1,171 @@
...
+    $ swift list
+    private-container
+    ```
+
+To retrieve more detailed information about an individual container,
+you can also use this command:
+
+=== "OpenStack CLI"
+    ```console
+    $ openstack container show test-container
```
mrtmm, line 56: I believe this should be `private-container`, right?

fghaas: Indeed it should! Good catch; fixed now.

mrtmm (approved):
Looks good to me now, thanks!

PR #83, opened 2022-12-07
feat: Add HowTo rescue server

ahosam:
This guide goes through the steps to rescue a server, this includes, listing rescue image options, performing a rescue, access server on rescue mode, Both through the GUI and OS CLI.


fghaas:
@ahosam Before we start reviewing this, please add a proper commit message body, and update your PR description. Do not refer to internal ticket references in doing so; this is a public project that we want to keep free of internal-only references.

docs/howto/openstack/nova/rescue-server.md
```diff
@@ -0,0 +1,89 @@
...
+    To start the rescue using system-rescue image use the following command:
+
+    ```bash
+    for uuid in `openstack image list --tag system-rescue -c ID -f value`; do openstack --os-compute-api-version 2.87 server rescue --image $uuid <server_id> ; done
+    ```
+
+    To start the rescue using server base image use the following command:
+
+    ```bash
+    openstack server rescue <server_id>
```
noonedeadpunk, line 65: I think this method should fail for BFV instances. We should be explicit on when to use rescue image. Given that these docs are not actually considering non-BFV, not sure we should even mention this (unless I'm wrong and bfv instances can boot with base image)

fghaas (changes requested):
Please remove everything in the `site` directory from your commit. All that content is automatically generated by the CI platform, and does not belong in the repo.

fghaas:
@ahosam I've taken the liberty to give this a rewrite, and @colder-is-better and I will now proceed with the review.

colder-is-better (commented):
Besides the fact that I'm not a big fan of taking screenshots of small graphical elements, like the three-dot icon or the logo-with-exclamation-mark icon, everything else looks good to me :)

fghaas (approved):
OK, then let's go ahead and merge this now. :)

noonedeadpunk:
You're so fast @fghaas :-)

I was going to say, that we'd need to mention on how to upload own rescue image in terms of properties that should be set for such image to work nicely, such as
```
hw_rescue_bus=virtio
hw_rescue_device=disk 
```

We also tag such images as `system-rescue` for purpose of this doc (being able to filter) and I expect CCP to implement filtering on this tag as well in the future.

I can't propose follow up patch right now due to ENOTIME, so have to ask you or @ahosam for adding section of adding own rescue images.

fghaas:
@noonedeadpunk Do we want to encourage that though? As in, do we want to suggest that people use a different rescue image from the one we test make available publicly?

noonedeadpunk:
@fghaas I'm not sure that systemrescuecd covers all use-cases, let's say, for Windows recovery. As  I can imagine that ppl might want some specific recovery image that we don't have (or don't want to have) at the moment.
Also systemrescuecd used to have quite tricky way for installing additional tools that were needed occasionally (but quite rarely). They've switched underlying OS from Gentoo to smth else since then (haven't used it a lot lately), so likely it's not the case as of today.
But still I don't see an issue of having custom recovery images if clients are willing so. And this enables CE to point to docs if they're unhappy with what we have.

PR #86, opened 2022-12-19
feat: Explain how to disable object expiry in the S3 how-to

fghaas:
Previously, we only explained how to enable object expiry on an
S3 bucket. Extend the how-to guide to also cover removing the
object expiry lifecycle policy.


colder-is-better (commented):
Everything LGTM!

fghaas:
> Everything LGTM!

OK, excellent. I just rebased on current `main` after merging #83. Please approve at your convenience, so we can merge this one too.

colder-is-better (approved):

PR #88, opened 2022-12-20
feat: Enable social cards

fghaas:
It sometimes comes in handy to share a bit of documentation on social
media (or chat), and the social cards plugin enables mkdocs-material
to generate some fancy cards for that purpose.

The cards themselves are not particularly useful without a brief
summary of the page being shared, and mkdocs-material neither has a
facility to auto-generate short page summaries, nor a provision to
mark up part of the page content as the summary.

However, we can apply a flipped approach where we

1. set the page description (which goes on the social card) in the YAML
   front matter for a page,
2. Use the macros plugin to then incorporate the description into the
page body.

This is something we can obviously gradually add over time, so apply
this approach to just a handful of pages for the time being.

One caveat does apply here, which is that we cannot use macro
expansion *within* the YAML front matter, so we cannot use things like
the {{brand}} placeholder in a description.

References:
https://squidfunk.github.io/mkdocs-material/setup/setting-up-social-cards/
https://squidfunk.github.io/mkdocs-material/reference/#setting-the-page-description


fghaas:
@colder-is-better This is just an idea, hence the draft PR. But I'd appreciate your thoughts — do you think this would be a good feature to add?

fghaas:
@colder-is-better Gotcha, thanks for the feedback! Do you think that the implementation is sensible too (within the constraints of the available functionality, that is)?

colder-is-better:
Even though the implementation cannot be automated, every time you have your post ready it won't be too much trouble to add a short description/summary. Also, the fact that you can't use placeholders is something that would never bother me.

fghaas:
OK! In that case let's apply this, if even only as a test balloon. We can always back it out if it doesn't work out, or adds too much overhead work.

If you're good with the changes at hand, please approve at your convenience.

colder-is-better (approved):

PR #93, opened 2023-01-03
Update linter configuration, facilitate one-sentence-per-line policy

fghaas:
Implement a few changes in the Markdown linter:

* Set the maximum line length for non-code-blocks and non-headings to 200 characters.
* Set the maximum line length for headings to 80 characters
* Disable line length checking for code blocks.
    
The rationale behind this is that it enables us to adopt a Markdown style with one sentence per line. This style has the following benefits:
    
* It facilitates code reviews, as it ensures that any change to a sentence is also a change to a line, and vice versa.
* It makes long run-on sentences stand out visually while editing, encouraging the author to break them into shorter and more readable sentences.
* It has no bearing on the flow of the rendered text, since the Markdown renderer will only care about line breaks preceded by two spaces (which become a line break within a paragraph), and about multiple consecutive line breaks (which mark the end of a paragraph).

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -1,32 +1,25 @@
 # Modifying content on this site
 
-You have two options for editing content: directly in your browser
-using GitHub, or using a Git-based workflow from your local work
-environment.
+You have two options for editing content: directly in your browser using GitHub, or using a Git-based workflow from your local work environment.
 
 ## Notes on content additions
 
+**Sections:** Generally, content additions should fit somewhere within the existing top-level and second-level sections (like [_Background_](../background/index.md), or [_Kubernetes_](../howto/kubernetes/index.md)).
+Try not to introduce a new top-level or second-level section.
 
-**Sections:** Generally, content additions should fit somewhere within
-the existing top-level and second-level sections. Try not to introduce
-a new top-level or second-level section.
+**Text:** Write your Markdown text in such a way that each sentence in a paragraph has its own line, and is followed immediately by a newline character with no spaces in between.
+This facilitates the review of your change, and it also helps you notice long run-on sentences as you write.
 
-**Screenshots:** If you are contributing a change that contains
-screenshots from {{gui}}, they should use a resolution of
-1920×1080 pixels (1080p). If your screen uses a larger resolution, use
-[Firefox Responsive Design
-Mode](https://firefox-source-docs.mozilla.org/devtools-user/responsive_design_mode/)
-or [Chrome/Chromium Device
-Mode](https://developer.chrome.com/docs/devtools/device-mode/) to
-configure your browser with 1080p.
-Screenshots should be added to a directory named `assets`, located in
-the same directory as the Markdown file you are adding or editing.
+**Headings:** Keep headings concise, under 80 characters.
 
-**CLI screen dumps:** If you are contributing a change that contains a
-screen dump from the `openstack` command-line client, please limit its
-width to 100 characters. You can do this by setting the following
-environment variable in your terminal, before you start work on your
-change.
+**Screenshots:** If you are contributing a change that contains screenshots from {{gui}}, they should use a resolution of 1920×1080 pixels (1080p).
+If your screen uses a larger resolution, use [Firefox Responsive Design Mode](https://firefox-source-docs.mozilla.org/devtools-user/responsive_design_mode/)
+or [Chrome/Chromium Device Mode](https://developer.chrome.com/docs/devtools/device-mode/) to configure your browser with 1080p.
+
+Screenshots should be added to a directory named `assets`, located in the same directory as the Markdown file you are adding or editing.
+
+**CLI screen dumps:** If you are contributing a change that contains a screen dump from the `openstack` command-line client, please limit its width to 100 characters.
+You can do this by setting the following environment variable in your terminal, before you start work on your change.
 
 ``` bash
 export CLIFF_MAX_TERM_WIDTH=100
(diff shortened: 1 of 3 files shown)


fghaas:
@colder-is-better This is an idea/proposal, hence the Draft state. I've recently come across the Markdown style of putting each sentence on its own line, and it is making a lot of sense to me for the reasons enumerated in the PR description. Would you like to give this PR a read, and let me know whether this sounds reasonable to you?

colder-is-better:
The reasons enumerated in the PR all make perfect sense, I don't doubt that. Personally, though, when I write, I prefer to focus on the content and not so much on rules like "one sentence per line". But maybe switching markdown editors and configuring the new one accordingly will allow me to do both (that is, focus on the content and follow the rules at the same time). So, let's go ahead with the new linter rules and see how those will work in practice.

fghaas:
@colder-is-better Do note that the changes as-is do not _enforce_ that rule (I don't see how we even could), they merely encourage the practice. Still, I'll do some more thinking on this.

docs/howto/openstack/nova/server-group.md
```diff
@@ -68,9 +72,11 @@ $ openstack server show -c fault -c status <server_id>
 +--------+--------------------------------------------------+
 ```
 
-This is normal as {{brand}} cannot schedule the server on a different physical compute node because there already is a server in the server group on every node. The same scheduling error and "fault" message will occur when using a server group with a policy of `affinity` as well, when you create more servers than a physical compute node can host.
+This is normal as {{brand}} cannot schedule the server on a different physical compute node because there already is a server in the server group on every node.
```
colder-is-better, line 75: Sounds a bit better inside muh head.

```suggestion
This is normal as {{brand}} cannot schedule the server on a different physical compute node because there is already a server in the server group on every node.
```

docs/contrib/modifications.md
```diff
@@ -87,37 +69,31 @@ Please see our [notes on commit messages](quality.md).
...
 
-A local copy of the documentation will then run on your local machine
-and be accessible from <http://localhost:8000> in your browser.
+A local copy of the documentation will then run on your local machine and be accessible from <http://localhost:8000> in your browser.
 
-When you are planning to make several changes in rapid succession, you
-may want to speed up rendering the site after each change. You may do
-so by disabling a plugin that checks all links (including external
-links) for accessibility:
+When you are planning to make several changes in rapid succession, you may want to speed up rendering the site after each change.
```
colder-is-better, line 95: A bit better -- at least to me :|

```suggestion
When planning to make several changes in rapid succession, you may want to speed up site rendering after each change.
```

docs/contrib/modifications.md
```diff
@@ -87,37 +69,31 @@ Please see our [notes on commit messages](quality.md).
...
 
 ```bash
 cd {{ brand | lower }}-docs
 git checkout <your-topic-branch-name>
 tox -e serve
 ```
 
-A local copy of the documentation will then run on your local machine
-and be accessible from <http://localhost:8000> in your browser.
+A local copy of the documentation will then run on your local machine and be accessible from <http://localhost:8000> in your browser.
```
colder-is-better, line 93: I did not like the first "local" much, so I came up with this suggestion:

```suggestion
A copy of the documentation will then be hosted on your local machine and be accessible from <http://localhost:8000> in your browser.
```

docs/contrib/modifications.md
```diff
@@ -1,66 +1,48 @@
...
-Every page on this site has an Edit button
-([:{{config.theme.icon.edit | replace('/','-')}}:]({{page.edit_url}})).
-If you click it, it’ll
-take you straight to its [corresponding source page]({{page.edit_url}}) in GitHub. Then,
-you can follow [GitHub’s
-documentation](https://docs.github.com/en/repositories/working-with-files/managing-files/editing-files#editing-files-in-another-users-repository)
-on how to propose changes to another user’s repository.
-
+Every page on this site has an Edit button ([:{{config.theme.icon.edit | replace('/','-')}}:]({{page.edit_url}})).
+If you click it, it’ll take you straight to its [corresponding source page]({{page.edit_url}}) in GitHub.
```
colder-is-better, line 31: "its" does not refer to the button (duh!), so I am thinking this instead:
```suggestion
If you click it, it’ll take you straight to the [corresponding source page]({{page.edit_url}}) in GitHub.
```

docs/contrib/modifications.md
```diff
@@ -1,66 +1,48 @@
...
-the same directory as the Markdown file you are adding or editing.
+**Headings:** Keep headings concise, under 80 characters.
 
-**CLI screen dumps:** If you are contributing a change that contains a
-screen dump from the `openstack` command-line client, please limit its
-width to 100 characters. You can do this by setting the following
-environment variable in your terminal, before you start work on your
-change.
+**Screenshots:** If you are contributing a change that contains screenshots from {{gui}}, they should use a resolution of 1920×1080 pixels (1080p).
+If your screen uses a larger resolution, use [Firefox Responsive Design Mode](https://firefox-source-docs.mozilla.org/devtools-user/responsive_design_mode/)
```
colder-is-better, line 16:
```suggestion
If your screen has a larger resolution, use [Firefox Responsive Design Mode](https://firefox-source-docs.mozilla.org/devtools-user/responsive_design_mode/)
```

docs/contrib/modifications.md
```diff
@@ -1,66 +1,48 @@
...
+or [Chrome/Chromium Device Mode](https://developer.chrome.com/docs/devtools/device-mode/) to configure your browser with 1080p.
+
+Screenshots should be added to a directory named `assets`, located in the same directory as the Markdown file you are adding or editing.
+
+**CLI screen dumps:** If you are contributing a change that contains a screen dump from the `openstack` command-line client, please limit its width to 100 characters.
+You can do this by setting the following environment variable in your terminal, before you start work on your change.
```
colder-is-better, line 22:
```suggestion
You can do this by setting the following environment variable in your terminal, before you start working on your change.
```

docs/contrib/modifications.md
```diff
@@ -1,66 +1,48 @@
 # Modifying content on this site
 
-You have two options for editing content: directly in your browser
-using GitHub, or using a Git-based workflow from your local work
-environment.
+You have two options for editing content: directly in your browser using GitHub, or using a Git-based workflow from your local work environment.
```
colder-is-better, line 3: This syntax is more consistent, IMHO:

```suggestion
You have two options for editing content: directly in your browser using GitHub, or from your local work environment using a Git-based workflow.
```

colder-is-better (commented):
All is looking good -- I have some non-planet-killing suggestions, though.

fghaas:
I appreciate the review, even though the comments had little to do with the stated purpose of the PR. :) I've now applied the suggested style fixes in 14084b36488136b58407da544bf2f4da70cd240b.

colder-is-better:
> I appreciate the review, even though the comments had little to do with the stated purpose of the PR. :)

Ah, yes, but I think we have already discussed that. In any case, reading about the purpose of the PR once more, I can say that I'm still of the mindset described in https://github.com/citynetwork/docs/pull/93#issuecomment-1369764926.

colder-is-better:
> The exact same functional change, applied to a text using filled paragraphs, results in three lines being changed, complicating review, auto-merge, and rebase.

Oh, but I agree with all of those technical aspects. What can I say, I am going to give this one-sentence-per-line policy a try. I very well know in advance that muh eyes will hurt while reading from the text editor, but at the same time, I also know I can survive a couple (more than two) of test runs.

fghaas:
@colder-is-better Okay, so do we merge this now? If in favour, please add your approval.

colder-is-better (approved):

PR #94, opened 2023-01-03
feat: Add how-to guide for retyping volumes

fghaas:
Add a how-to guide for changing the type of a volume.

### DIFF
--- a/docs/howto/openstack/cinder/retype-volumes.md
+++ b/docs/howto/openstack/cinder/retype-volumes.md
@@ -1,0 +1,134 @@
+---
+description: Changing the type of a volume ("retyping") is an offline operation that requires detaching the volume from its server, and setting its new type.
+---
+# Changing a volume's type
+
+You may occasionally need to change the type of a volume.
+This may be due to a volume type being phased out on {{brand}}'s part, necessitating data migration.
+Or you might want to enable or disable [volume-level encryption](encrypted-volumes.md).
+
+{{page.meta.description}}
+The resulting downtime may be quite substantial, particularly for large volumes.
+As such, when you need to retype volumes, you should plan ahead well in advance.
+
+## Prerequisites
+
+In order to retype volumes, you must use the OpenStack CLI, so make sure you [have it enabled](../../getting-started/enable-openstack-cli.md).
+
+If you are about to retype a large volume, or one that holds data associated with a critical service, you may be interested in an estimate of how long the retype operation will take.
+In that case, please file a support request with our [{{support}}](https://{{support_domain}}).
+
+## Checking the volume's state
+
+Assume you have a volume named `testvol` that is currently attached to a server named `testsrv`:
+
+```console
+$ openstack volume list --long
++----------------+---------+--------+------+---------+----------+----------------+------------+
+| ID             | Name    | Status | Size | Type    | Bootable | Attached to    | Properties |
++----------------+---------+--------+------+---------+----------+----------------+------------+
+| e233e7f3-f33b- | testvol | in-use |   50 | default | false    | Attached to    |            |
+| 4d7a-8f5b-785b |         |        |      |         |          | testsrv on     |            |
+| 34f670bf       |         |        |      |         |          | /dev/vdb       |            |
++----------------+---------+--------+------+---------+----------+----------------+------------+
+```
+
+In this example, this volume status is `in-use`, meaning it is currently attached to a server, and the volume type is `default`.
+
+## Detaching the volume
+
+You cannot retype a volume while it is attached to a server.
+You must thus detach it first.
+It is safest to do this while the server is shut down:
+
+```console
+$ openstack server stop testsrv
+
+$ openstack server list -c Name -c Status
++---------+---------+
+| Name    | Status  |
++---------+---------+
+| testsrv | SHUTOFF |
++---------+---------+
+```
+
+Once your server is in the `SHUTOFF` state, you can safely proceed to detaching the volume.
+This will change the volume status from `in-use` to `available`.
+
+```console
+$ openstack server remove volume testsrv testvol
+
+$ openstack volume list
++--------------------------------------+---------+-----------+------+-------------+
+| ID                                   | Name    | Status    | Size | Attached to |
++--------------------------------------+---------+-----------+------+-------------+
+| e233e7f3-f33b-4d7a-8f5b-785b34f670bf | testvol | available |   50 |             |
++--------------------------------------+---------+-----------+------+-------------+
+```
+
+## Retyping the volume
+
+With the volume safely detached, you can now change its volume type.
+If you were to change the volume type from `default` to `ceph_hdd`, you would proceed as follows:
+
+```console
+$ openstack volume set --type ceph_hdd --retype-policy on-demand testvol
+
+$ openstack volume list --long
++---------------+---------+----------+------+---------+----------+-------------+------------+
+| ID            | Name    | Status   | Size | Type    | Bootable | Attached to | Properties |
++---------------+---------+----------+------+---------+----------+-------------+------------+
+| e233e7f3-f33b | testvol | retyping |   50 | default | false    |             |            |
+| -4d7a-8f5b-78 |         |          |      |         |          |             |            |
+| 5b34f670bf    |         |          |      |         |          |             |            |
++---------------+---------+----------+------+---------+----------+-------------+------------+
+```
+
+Note that the volume status changes from `available` to `retyping`: this status change kicks off the actual data migration, which might take a significant amount of time.
+
+## Re-attaching the volume
+
+If you attempt to re-attach the volume while it is still retyping, you will receive an error:
+
+```console
+$ openstack server add volume testsrv testvol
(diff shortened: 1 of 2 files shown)


docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,106 @@
...
+| ID            | Name    | Status    | Size | Type     | Bootable | Attached to | Properties |
++---------------+---------+-----------+------+----------+----------+-------------+------------+
+| d2a22868-a133 | testvol | available |    2 | ceph_hdd | false    |             |            |
+| -40a1-b1a6-0c |         |           |      |          |          |             |            |
+| bae3feaf8d    |         |           |      |          |          |             |            |
++---------------+---------+-----------+------+----------+----------+-------------+------------+
+```
+
+It is at this stage that the actual data migration occurs.
+As a result, this step might take a significant amount of time.
```
fghaas, line 87: @noonedeadpunk, in this section we're still missing the bit where the user should check on whether the migration has completed (i.e. where I don't know if they're even able to, without admin privileges).

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
+---
+description: Changing the type of a volume ("retyping") is an offline operation that requires detaching the volume from its server, and setting its new type.
+---
+# Changing a volume's type
+
+You may occasionally need to change the type of a volume.
+This may be due to a volume type being phased out on {{brand}}'s part (necessitating data migration), or you might want to enable or disable [encryption](encrypted-volumes.md) on a volume.
```
colder-is-better, line 7: Instead of "you might want to enable or disable [encryption](encrypted-volumes.md) on a volume" I would write "you might want to enable or disable [encryption](encrypted-volumes.md)" or "you might want to enable or disable volume-level [encryption](encrypted-volumes.md)".

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
+
+{{page.meta.description}}
+The resulting down time may be quite substantial, particularly for large volumes.
```
colder-is-better, line 10: Instead of "down time," I would write "downtime" (that's a case where a run-together term is preferred).

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
+As such, when you need to retype volumes, you should plan ahead well in advance.
+
+## Prerequisites
+
+In order to retype volumes, you must use the OpenStack CLI.
+Make sure you have [enabled it](../../getting-started/enable-openstack-cli.md).
+
```
colder-is-better, line 17: OK, this is just me, but I feel this is more "natural":

"In order to retype volumes you must use the OpenStack CLI, so make sure you have it [enabled](../../getting-started/enable-openstack-cli.md)"

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
++----------------+---------+--------+------+---------+----------+----------------+------------+
+| ID             | Name    | Status | Size | Type    | Bootable | Attached to    | Properties |
++----------------+---------+--------+------+---------+----------+----------------+------------+
+| e233e7f3-f33b- | testvol | in-use |   50 | default | false    | Attached to    |            |
+| 4d7a-8f5b-785b |         |        |      |         |          | testsrv on     |            |
+| 34f670bf       |         |        |      |         |          | /dev/vdb       |            |
++----------------+---------+--------+------+---------+----------+----------------+------------+
+```
+
+In this example, this volume's status is `in-use`, meaning it is currently attached to a server, and the volume's type is `default`.
```
colder-is-better, line 36: I would not use the possessive form --- not for the volume, at least. So, I would write something like this: "In this example, the volume status is `in-use`, meaning it is currently attached to a server and that the volume type is `default`".

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
+$ openstack server list -c Name -c Status
++---------+---------+
+| Name    | Status  |
++---------+---------+
+| testsrv | SHUTOFF |
++---------+---------+
+```
+
+Once your server is in the `SHUTOFF` state, you can safely proceed to detaching the volume.
+This will change the volume's status from `in-use` to `available`.
```
colder-is-better, line 56: Uhm, "volume status" instead of "volume's status", please?

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
+| ID                                   | Name    | Status    | Size | Attached to |
++--------------------------------------+---------+-----------+------+-------------+
+| e233e7f3-f33b-4d7a-8f5b-785b34f670bf | testvol | available |   50 |             |
++--------------------------------------+---------+-----------+------+-------------+
+```
+
+## Retyping the volume
+
+With the volume safely detached, you can now change its volume type.
+If you want to change the volume type from `default` to `ceph_hdd`, you would proceed as follows:
```
colder-is-better, line 72: Instead of "If you want to change", maybe write "If, for instance, you want to change"?

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,135 @@
...
++---------------+---------+----------+------+---------+----------+-------------+------------+
+| ID            | Name    | Status   | Size | Type    | Bootable | Attached to | Properties |
++---------------+---------+----------+------+---------+----------+-------------+------------+
+| e233e7f3-f33b | testvol | retyping |   50 | default | false    |             |            |
+| -4d7a-8f5b-78 |         |          |      |         |          |             |            |
+| 5b34f670bf    |         |          |      |         |          |             |            |
++---------------+---------+----------+------+---------+----------+-------------+------------+
+```
+
+Note that the volume's status changes from `available` to `retyping`: it is at this stage that the actual data migration occurs.
```
colder-is-better, line 87: (cough) Possessive form alert (cough)

colder-is-better (commented):
All is looking good --- I have only some quantum-level nits.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: OK.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: Don't mind either way. Will change.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: I think it's a toss-up, but yeah we can go without the possessive.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: If you insist.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: I'll put "if you were to change" which is just as speculative but shorter.

fghaas:
@colder-is-better I *think* this one is good to go — if you agree, would you mind approving the PR?

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,134 @@
...
+| 34f670bf       |         |        |      |         |          | /dev/vdb       |            |
++----------------+---------+--------+------+---------+----------+----------------+------------+
+```
+
+In this example, this volume status is `in-use`, meaning it is currently attached to a server, and the volume type is `default`.
+
+## Detaching the volume
+
+You cannot retype a volume while it is attached to a server.
+You must thus detach it first.
```
noonedeadpunk, line 41: Should we say that root volumes can't be detached?

docs/howto/openstack/cinder/retype-volumes.md
```diff
@@ -0,0 +1,134 @@
...
+```console
+$ openstack server add volume testsrv testvol
+BadRequestException: 400: Client Error for url: https://sto2.{{brand_domain}}:8774/v2.1/servers/b6a26b0e-f911-4ea2-8e45-51f16442da03/os-volume_attachments, Invalid input received: Invalid volume: Volume e233e7f3-f33b-4d7a-8f5b-785b34f670bf status must be available or downloading to reserve, but the current status is retyping. (HTTP 400) (Request-ID: req-36c4f2e5-ef2f-4ff4-b912-0baed2594f4d)
+```
+
+You must now wait until the volume status changes back from `retyping` to `available`.
+One way to do this is with a bash `until` loop:
+
+```bash
+until [ `openstack volume show -f value -c status testvol` = "available" ]; do
```
noonedeadpunk, line 102: Maybe it would be more convenient for users to have that as oneliner? that's more readable from other side.

docs/howto/openstack/cinder/retype-volumes.md
fghaas: Please feel free to send a longer patch, _after_ this one lands, explaining how to deal with BFV and retyping. That probably belongs in the Background section too, not in a how-to guide.

colder-is-better (approved):
Happily approving :)

PR #97, opened 2023-01-04
refactor: Introduce macros for the REST API and its domain

fghaas:
Just to ensure that we are consistent across pages, and to facilitate
any future changes to the Cleura Cloud REST API name or the domain
where we host it, introduce macro variables for both of them.


colder-is-better (approved):
That's nice and looks good!

PR #98, opened 2023-01-04
docs: Add an issue template for missing or incomplete documentation

fghaas:
Up to this point we have only allowed bug reports, because the site was known to be very incomplete and we had a long list of things to add before we could even consider it useful.

Now, we are getting much closer to that stage, and once we are satisfied that we can take on documentation requests, we should be ready to take them on.

Since this is GitHub specific, and the issue template can only be meaningfully tested if this change lands in main, we merge it now, test it, and then back it out again to be brought back once we drop the "beta" moniker for the site.

This uses a YAML issue template, rather than the Markdown ones we have been using for bug reports thus far.

Reference:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository

### DIFF
--- a/.github/ISSUE_TEMPLATE/gap.yml
+++ b/.github/ISSUE_TEMPLATE/gap.yml
@@ -1,0 +1,62 @@
+---
+about: >
+  Use this template to report documentation gaps: information that is
+  missing altogether, or that is incomplete.
+assignees: ''
+body:
+  - attributes:
+      label: Summary
+      placeholder: >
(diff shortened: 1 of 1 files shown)


colder-is-better (commented):
LGTM!

fghaas:
@colder-is-better OK great! In that case, please approve so I can merge this.

colder-is-better (approved):

PR #99, opened 2023-01-04
fix: Correct YAML syntax for issue template

fghaas:
Fix the following issues by which the `gap.yml` issue template was invalid as per the applicable schema:

* Remove the `title`, `assignees`, `labels` keys (which cannot be blank if set, and it does not make sense to preseed a value for them).
* Remove the `about` key, replace it with `description`.
* Fix typo in `validations` (was `validation`).

References:
* https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms 
* https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-githubs-form-schema

### DIFF
--- a/.github/ISSUE_TEMPLATE/gap.yml
+++ b/.github/ISSUE_TEMPLATE/gap.yml
@@ -1,8 +1,4 @@
 ---
-about: >
-  Use this template to report documentation gaps: information that is
-  missing altogether, or that is incomplete.
-assignees: ''
 body:
   - attributes:
       label: Summary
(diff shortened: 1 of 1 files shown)


colder-is-better (dismissed):

fghaas:
Gnah, looks like you approved seconds before I pushed one more update. Sorry about that! Please re-approve at your convenience. Thanks!

colder-is-better (approved):

PR #100, opened 2023-01-04
build: Use tox-gh-actions as intended

fghaas:
Rather than enumerating the tox testenvs with `tox -e` from the `build.yml` workflow, add a `[gh-actions]` section to `tox.ini` and run just `tox` from the workflow.


fghaas:
This just doesn't work. For some reason, `tox-gh-actions` just ignores the `[gh-actions]` section in `tox.ini`, and only runs the default envlist.

Superseded by #102.

PR #101, opened 2023-01-04
fix: Change URL textarea to input in issue template

fghaas:
It's a bit silly to use a multi-line textarea for a URL, so just
change it to a one-line input box.


colder-is-better (approved):
Aha, makes sense!

PR #102, opened 2023-01-04
build: Drop tox-gh-actions, simplify tox invocation from GitHub Actions

fghaas:
Since we are not using a `[gh-actions]` section in `tox.ini`, there is really no need to keep installing `tox-gh-actions`. Drop it.

In addition, rather that enumerating all the testenvs we want to run from our GitHub Actions workflow, just invoke `tox` twice: first check with only `gitlint`, then run the default testenvs specified in `tox.ini`'s `envlist`.


colder-is-better (approved):
Looking good!

PR #103, opened 2023-01-04
fix: Explain that the REST API only supports SMS 2FA

fghaas:
In the getting started guide for using the REST API, explain that accounts with only Webauthn 2FA are unsuitable for REST API use, and that instead users must configure an account with SMS as a second factor.

Also, specify that users should always specify the `twofa_method` parameter and set it to `"sms"`: if they have only SMS configured as a second factor, it won't hurt to use that parameter. If they have SMS _and_ Webauthn, they must set it (even though Webauthn doesn't work for getting a token).


colder-is-better (approved):
All changes make sense and cover every case.

PR #106, opened 2023-01-09
feat: Add how-to guide for transferring data between volumes

fghaas:
Add a how-to explaining how to move data from one volume to another,
using snapshots and (optionally) retyping.


colder-is-better (approved):
Very detailed --- and it sure LGTM!

PR #107, opened 2023-01-09
fix: Restore edit button

fghaas:
Our contribution guide mentions an edit button on every page. However, as of mkdocs-material 9, this edit button is no longer on by default if `repo_url` is set, but it is toggled by a feature flag. It also uses a different icon.

Enable the feature flag, and also update the contribution guide to refer to the updated icon. In doing so, use a bit of Jinja trickery with mkdocs-macros so that the in-text edit icon auto-updates with the one in the configuration (in case we ever change it).

Reference:
https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#code-actions

### DIFF
--- a/docs/contrib/modifications.md
+++ b/docs/contrib/modifications.md
@@ -34,8 +34,10 @@ export CLIFF_MAX_TERM_WIDTH=100
 
 ## Modifying content from your browser
 
-Every page on this site has an Edit button 🖍️. If you click it, it’ll
-take you straight to the corresponding source page in GitHub. Then,
+Every page on this site has an Edit button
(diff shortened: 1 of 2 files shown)


colder-is-better (commented):
Looking good! There's something weird in the commit message, though; there's an unexpected new line after "trickery" (which is somewhat ironic, now that I think about it).

fghaas:
> There's something weird in the commit message, though; there's an unexpected new line after "trickery" (which is somewhat ironic, now that I think about it).

That problem isn't with the commit message, but with the PR description (those being two different things, this being GitHub and not Gerrit). I've fixed the PR description now.

colder-is-better (approved):

PR #109, opened 2023-01-11
fix: Add missing .pages file in OpenStack API reference section

fghaas:
Fix the navigation so that it says "OpenStack API" in the API reference section, akin to "Cleura Cloud REST API".

(This file was accidentally omitted from #108.)


colder-is-better (approved):
Looking good!

PR #110, opened 2023-01-11
feat: Add how-to guide for creating a publicly readable S3 bucket

fghaas:
Explain how to use S3 bucket policies to make all objects in a bucket publicly accessible via a browser or `curl`, and optionally how to enable bucket listing.

### DIFF
--- a/docs/howto/object-storage/s3/public-bucket.md
+++ b/docs/howto/object-storage/s3/public-bucket.md
@@ -1,0 +1,128 @@
+---
+description: You can use the S3 API to configure a bucket with public read access, so that anyone can download its objects with a web browser.
+---
+# Public buckets
+
+{{page.meta.description}}
+Making a bucket globally readable entails setting a *bucket policy* that enables read access on all its objects.
+
+> Setting a bucket policy affects **all** objects in a bucket.
+> To avoid inadvertent disclosure of existing information, consider setting public read policies only on empty buckets.
+
+## Prerequisites
+
+Object versioning requires that you configure your environment with [working S3-compatible credentials](credentials.md).
+
+## Setting a public read policy for a bucket
+
+First, create a local policy file named `policy.json`, with the following content (replace `<bucket-name>` with the name of your bucket):
+
+```json
+{ 
+  "Statement": [ 
+    {
+      "Effect": "Allow",
+      "Principal": "*",
+      "Action": "s3:GetObject",
+      "Resource": "arn:aws:s3:::<bucket-name>/*"
+    }
+  ]
+}
+```
+
+To apply this policy to a bucket such that read-only access is permitted for everyone, run the following command:
+
+=== "aws"
+    ```bash
+    aws --profile <region> \
+      --endpoint-url=https://s3-<region>.{{brand_domain}}:8080 \
+      s3api put-bucket-policy \
+      --policy file://policy.json \
+      --bucket <bucket-name>
+    ```
+=== "mc"
+    ```bash
+    mc anonymous set-json policy.json <region>/<bucket-name>
+    ```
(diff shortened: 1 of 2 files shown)


docs/howto/object-storage/s3/public-bucket.md
```diff
@@ -0,0 +1,129 @@
...
+
+Once you have updated your local `policy.json` file, apply it just as when you created the policy.
+
+You can then open a bucket path in your browser, to retrieve an XML document containing a list of all objects in the bucket.
+You would construct the URL by the following schema:
+
+```plain
+https://s3-<region>.{{brand_domain}}:8080/<project-uuid>:<bucket-name>
+```
+
```
colder-is-better, line 109: Extra-line semi-alert :)

docs/howto/object-storage/s3/public-bucket.md
```diff
@@ -0,0 +1,129 @@
...
+                                 Dload  Upload   Total   Spent    Left  Speed
+100 62703  100 62703    0     0   186k      0 --:--:-- --:--:-- --:--:--  186k
+```
+
+## Enabling bucket listing
+
+The `policy.json` file above allows anyone to retrieve known objects by name, but does not enable listing the bucket's contents.
+Most of the time, this is what you want.
+
+However, in the event that you do need unauthenticated clients to be able to list all objects in a bucket, modify your `policy.json` file as in the following example:
```
colder-is-better, line 78: I would use a simpler syntax ("in the event" reminds me of the instructions of your friendly flight attendant): "However, if you do need unauthenticated clients".

docs/howto/object-storage/s3/public-bucket.md
```diff
@@ -0,0 +1,129 @@
...
+
+```plain
+https://s3-<region>.{{brand_domain}}:8080/<project-uuid>:<bucket-name>/object-name
+```
+
+> Your project UUID is listed as the `project_id` field in the output of the `openstack ec2 credentials create` command you used to [create your S3-compatible credentials](credentials.md).
+>
+> If you did not note it down at the time of account creation, you can always retrieve it with `openstack ec2 credentials <access-key>`.
+
+For example, to retrieve an object named `bar.pdf` in a bucket named `foo` that belonged to the project with the UUID `07576783684248f7b2745e34356c6025` in the {{brand}} Kna1 region, you would run:
```
colder-is-better, line 64: I would write something like this: "in a bucket named `foo` from a project with UUID"

colder-is-better (commented):
Looking good --- sans a couple of sub-atomic nits.

docs/howto/object-storage/s3/public-bucket.md
fghaas: Fair! :)

docs/howto/object-storage/s3/public-bucket.md
fghaas: I went with "in case" rather than "in the event that". That's more concise, but it still implies that it should be the exception, rather than the norm (whereas a simple "if" strikes me as completely neutral).

docs/howto/object-storage/s3/public-bucket.md
fghaas: I used "from the project with the UUID".

fghaas:
I appreciate you catching the nits! Patch updated, should hopefully be good to merge now. If you agree, please approve. :)

docs/howto/object-storage/s3/public-bucket.md
colder-is-better: Ah, yes, "in case" is more suitable!

colder-is-better (approved):

PR #111, opened 2023-01-12
build: Toggle link checker in Git hooks, disable PDF output

fghaas:
PDF exports of the documentation pages are currently broken in two ways:
    
* Content tabs no longer convert to subsections that are shown one after the other, like they used to.
* Fenced code blocks render in a proportional, rather than a monospace font.
    
This renders the PDF output essentially useless, so until we can fix it, just disable it.

Also, set/unset the dead link checker in Git hooks. It's usually tedious and time-consuming to run the dead link checker on every commit, particularly if one is working on a local topic branch that gets frequent commits. So, update the `pre-commit` hook script to disable it (unless it has been specifically enabled in the environment already). However, enable it unconditionally in the `pre-push` hook so that we don't create broken pull requests.

In addition, include the commit hooks in the list of scripts linted by `bashate`.

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,3 +1,7 @@
-#!/bin/sh
+#!/bin/bash
+
+if [ -z "${DOCS_ENABLE_HTMLPROOFER}" ]; then
+    export DOCS_ENABLE_HTMLPROOFER=false
+fi
 
 exec tox
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,3 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
+export DOCS_ENABLE_HTMLPROOFER=true
 exec tox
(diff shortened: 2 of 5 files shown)


.githooks/pre-commit
```diff
@@ -1,3 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 
-exec tox
\ No newline at end of file
+if [ -z "${DOCS_ENABLE_HTMLPROOFER}" ]; then
+    export DOCS_ENABLE_HTMLPROOFER=false
+fi
```
colder-is-better, line 5: I am wondering, wouldn't we save a couple of pico-seconds if we skipped the test altogether and just explicitly set `DOCS_ENABLE_HTMLPROOFER` to `false`?

colder-is-better (commented):
Looks good --- and I only have a question.

.githooks/pre-commit
fghaas: Saving picoseconds is irrelevant; what I want to be able to do is *enable* link checking on commit by saying

```bash
export DOCS_ENABLE_HTMLPROOFER=true
git commit
```

And if the test isn't there, I can't do that.

PR #112, opened 2023-01-12
feat: Add how-to guide for pre-signed URLs

fghaas:
Explain how to create pre-signed URLs. Also explain how the `Content-Disposition` header affects objects downloaded from pre-signed URLs.

### DIFF
--- a/docs/howto/object-storage/s3/presign.md
+++ b/docs/howto/object-storage/s3/presign.md
@@ -1,0 +1,166 @@
+---
+description: You can use the S3 API to temporarily allow access to individual objects without authentication from a browser, using pre-signed URLs.
+---
+# Pre-signed object URLs
+
+{{page.meta.description}}
+
+This approach is distinct from enabling *permanent* [anonymous access to objects in a bucket](public-bucket.md).
+If you are planning to use pre-signed URLs in a bucket, then that bucket should *not* have a public read policy.
+
+> Pre-signed URLs are typically found in web applications using a Software Development Kit (SDK) to interact with the S3 API.
+> An example of such an SDK would be [Boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/index.html) for Python applications.
+>
+> This how-to guide instead uses command-line utilities to illustrate the concept, and to facilitate testing pre-signed URLs in {{brand}}.
+
+## Prerequisites
+
+The use of pre-signed URLs requires that you configure your environment with [working S3-compatible credentials](credentials.md).
+
+## Creating a pre-signed URL
+
+To create a pre-signed URL for an object, you use the following command (replace `<seconds>` with the number of seconds you want the pre-signed URL to be valid):
+
+=== "aws"
+    ```bash
+    aws --profile <region> \
+      --endpoint-url=https://s3-<region>.{{brand_domain}}:8080 \
+      s3 presign \
+      --expires-in <seconds>
+      s3://<bucket-name>/<object-name>
+    ```
+=== "mc"
+    ```bash
+    mc share download --expire <seconds>s <region>/<bucket-name>
+    ```
+    Note the `s` suffix for the `--expire` option.
+    `mc` also supports the suffix `m` for minutes, `h` for hours, and `d` for days.
+
+    If you do not set `--expire`, `mc` defaults to creating a pre-signed URL that is valid for 7 days.
+=== "s3cmd"
+    > For pre-signed URLs generated with `s3cmd` to work correctly in {{brand}}, you must use an `s3cmd` version later than 2.0.1, and set `signurl_use_https = True` in your configuration file.
+
+    ```bash
+    s3cmd -c ~/.s3cfg-<region> signurl s3://<bucket-name>/<object-name> +<seconds>
+    ```
+    Note the `+` prefix on the expiry.
+    `s3cmd` also supports setting an absolute expiry date, which does not use the `+` prefix and must be formatted in [Unix time](https://en.wikipedia.org/wiki/Unix_time).
+
+The command will return the valid pre-signed URL.
+
+## Accessing objects with a pre-signed URL
+
+To access an object in a public bucket from a web browser or a generic HTTP/HTTPS client like `curl`, open the URL that the pre-sign command returned:
+
+```console
+curl -f -O https://s3-<region>.{{brand_domain}}:8080/<bucket-name>/<object-name>?AWSAccessKeyId=<access-key>&Signature=<signature>&Expires=<expiry>
+```
+
+As long as the query parameters are correct and the signature has not yet expired, this command will succeed.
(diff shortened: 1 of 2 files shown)


docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+
+    ```bash
+    s3cmd -c ~/.s3cfg-<region> signurl s3://<bucket-name>/<object-name> +<seconds>
+    ```
+    Note the `+` prefix on the expiry.
+    `s3cmd` also supports setting an absolute expiry date, which does not use the `+` prefix and must be formatted in [Unix time](https://en.wikipedia.org/wiki/Unix_time).
+
+The command will return the valid pre-signed URL.
+
+
```
colder-is-better, line 51: Just an extra empty line.

docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+## Accessing objects with a pre-signed URL
+
+To access an object in a public bucket from a web browser or a generic HTTP/HTTPS client like `curl`, open the URL that the pre-sign command returned:
+
+```console
+curl -f -O https://s3-<region>.{{brand_domain}}:8080/<bucket-name>/<object-name>?AWSAccessKeyId=<access-key>&Signature=<signature>&Expires=<expiry>
+```
+
+As long as the query parameters are correct and the signature has not yet expired, this command will succeed.
+If the query parameters are incorrect or the pre-signed URL is past its expiry date, it will fail with HTTP 403 (Forbidden) instead.
```
colder-is-better, line 61: Instead of "past its expiry date" I would write "past its expiration date".

docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+For example, to retrieve an object named `bar.pdf` in a bucket named `foo` in the {{brand}} Kna1 region via its pre-signed URL, you would run:
+
+```console
+$ curl -o bar.pdf https://s3-kna1.{{brand_domain}}:8080/foo/bar.pdf?AWSAccessKeyId=07576783684248f7b2745e34356c6025&Expires=1673521496&Signature=%2Frm9nLV3moP%2FQz7aGCAnrESXjbk%3D
+  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
+                                 Dload  Upload   Total   Spent    Left  Speed
+100 62703  100 62703    0     0   186k      0 --:--:-- --:--:-- --:--:--  186k
+```
+
+## Setting the download file name on the object
```
colder-is-better, line 75: Uhm, "filename" instead of "file name"? (My preference, at least.)

docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+To ensure that an object named `bar.pdf` in a bucket named `foo` is always downloaded and stored with `bar.pdf` as its filename, you can set its `Content-Disposition` header.
+
+### Setting `Content-Disposition` on object creation
+
+=== "aws"
+    ```bash
+    aws --profile <region> \
+      --endpoint-url=https://s3-<region>.{{brand_domain}}:8080 \
+      s3api put-object \
+      --content-disposition 'attachment;filename="bar.pdf" \
```
colder-is-better, line 89: I see the left `'` but not the right `'` :/

docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+### Setting `Content-Disposition` on existing objects
+
+=== "aws"
+    To modify the `Content-Disposition` header of an existing object without downloading and re-uploading its contents, you must use `s3api copy-object`, with the object being its own copy source, and the metadata directive set to `replace`:
+    ```bash
+    aws --profile <region> \
+      --endpoint-url=https://s3-<region>.{{brand_domain}}:8080 \
+      s3api copy-object \
+      --copy-source <bucket-name>/<object-name>
+      --content-disposition 'attachment;filename="bar.pdf" \
```
colder-is-better, line 130: Same here, I do not see the right `'`.

docs/howto/object-storage/s3/presign.md
```diff
@@ -0,0 +1,167 @@
...
+
+Please keep in mind that there is nothing that keeps you from having two objects in the same bucket, with `Content-Disposition` on one object matching the object name of the other.
+
+For example, in the `foo` bucket, you may have
+
+* an object named `spam.pdf` with no `Content-Disposition` header, and
+* another object named `eggs.pdf`, with `Content-Disposition` set to include `filename="spam.pdf"`.
+
+If you were now to generate a pre-signed URL on `eggs.pdf` and downloaded it with `curl -OJ`, you would end up with a local file *named* `spam.pdf`, with the *contents* of `eggs.pdf`.
+This is obviously **not a good idea,** and you should avoid setting a `Content-Disposition` header that does not agree with the object name.
```
colder-is-better, line 167: Maybe the comma should not be in boldface? (Please!)

colder-is-better (commented):
All looks fine to me, besides a couple of infinitesimally small pico-nits.

docs/howto/object-storage/s3/presign.md
fghaas: Oh **nice catch!** I am impressed.

docs/howto/object-storage/s3/presign.md
fghaas: So, the way I've been taught, if something is in bold or italics immediately preceding punctuation, then the punctuation is included in that emphasis. Is that wrong?

docs/howto/object-storage/s3/presign.md
fghaas: I actually polled native speakers in my circle about this very issue a few months ago, and the preference for "expiry" vs. "expiration" is entirely regional. Since we are not following a particular national standard of English here, I prefer to be concise and use the shorter word.

docs/howto/object-storage/s3/presign.md
fghaas: Ack.

docs/howto/object-storage/s3/presign.md
colder-is-better: Well, I had no idea there was a rule for that :/ Do not change the boldness, then!

docs/howto/object-storage/s3/presign.md
fghaas: Yes, I used "filename" everywhere else in the document too. So let's be consistent.

docs/howto/object-storage/s3/presign.md
colder-is-better: Fair enough!

colder-is-better (approved):

PR #113, opened 2023-01-13
refactor: Add reference URLs for Swift and S3

fghaas:
Rather than repetitively open-coding Swift and S3 URLs, define the
base URL as a macro and use that.

This also gives us the ability to change things in one swoop if the
endpoints were ever to change.


fghaas:
Thinko. Doesn't make sense after all.

PR #114, opened 2023-01-13
feat: Explain that Swift and S3 public URLs are interchangeable

fghaas:
It perhaps isn't immediately obvious that our object storage buckets
and objects are one and the same when accessed via alternative paths
to the S3 and Swift APIs, and that as a result once an object has been
made public via one API, it's also public via the other.

Thus, add cross-references between the corresponding S3 and Swift
how-to guides explaining this.


colder-is-better (approved):
LGTM!

PR #115, opened 2023-01-16
fix: Explain that the volume sync doesn't apply to boot volumes

fghaas:
In the how-to guide on synchronizing volumes with minimal downtime, explain that the process does not apply for boot volumes (and suggest to retype those volumes offline, instead).


colder-is-better (dismissed):
LGTM!

fghaas:
Force-pushed this with no changes in order to rekick the CI pipeline, hoping that this time around the CI status is being reported correctly. Once the checks pass, we'll need a re-approval to be able to merge this.

colder-is-better (approved):

PR #116, opened 2023-01-17
feat: Add HOWTO: Create & test a TCP load balancer

colder-is-better:
We show how to create, configure, and test a TCP load balancer using the Cleura Cloud Control Panel or the OpenStack CLI. This is a detailed walk-through that guides the reader via a specific scenario.


fghaas (commented):
Overall this looks pretty good to me now, thanks! I have one minor and two medium-sized suggestions. :)

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,585 @@
...
+    ```text
+    +---------------------+--------------------------------------+
+    | Field               | Value                                |
+    +---------------------+--------------------------------------+
+    | address             | 10.15.25.105                         |
+    | admin_state_up      | True                                 |
+    | created_at          | 2023-01-16T20:11:08                  |
+    | id                  | 93aed0c8-b425-40b5-b166-f3e44ad9a986 |
+    | name                |                                      |
+    | operating_status    | NO_MONITOR                           |
```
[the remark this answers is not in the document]
colder-is-better: Ah, yes, I did not mention health monitoring at all. I will make sure to update the guide accordingly.

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,585 @@
...
+```bash
+ncat -kv -l 61234 -c 'echo Yello from $(hostname)!'
+```
+
+Since the load balancer has a floating IP (`198.51.100.129`,
+in our example). and we know the port it listens to, we can
+try connecting to it via `wget` and see what happens:
+
+```bash
+wget -q 198.51.100.129:61234 -O -
```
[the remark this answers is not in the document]
colder-is-better: Indeed, `curl` looks like the better choice. In our case, though, we need to add the `--http0.9` parameter. If we don't, we get this error: `curl: (1) Received HTTP/0.9 when not allowed`. While I am fine with adding the `--http0.9` option to `curl`, I suspect that the `wget` route feels more intuitive/natural/straightforward to the reader. What do you think? Should we go with `wget` or with `curl`?

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,585 @@
...
+```
+
+```plain
+Yello from srv-lbaas-2!
+```
+
+During our testing, executing the `wget` command repeatedly we were
+getting those two hostnames one after the other --- and that was
+proof our TCP load balancer was working as expected.
```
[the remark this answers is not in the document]
colder-is-better: Aha! That would greatly expand our test coverage. I will make sure to update the corresponding section accordingly.

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,585 @@
...
+    ```bash
+    pip install python-octaviaclient
+    ```
+
+## Scenario and terminology
+
+To configure and test our basic TCP load balancer, we have two servers
+in the same {{brand}} region, both having `ncat` listening on port
+61234/TCP and returning a greeting to clients asking to connect to that
+port. Both servers are in the same internal network, and our load
```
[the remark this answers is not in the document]
colder-is-better: Hmm, while it is correct to say, "I am on Facebook", "I'm on the network", etc., when we are talking about _things_ (not humans) that are actual parts of a network, then we use "in". This is my understanding --- and I am sure I often get this wrong.

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,585 @@
...
+balancer will be responsible for redirecting client connections to them
+in a round-robin fashion. The load balancer will have a single floating
+IP address, so prospective clients will have to know that IP and not
+any of the backend servers.
```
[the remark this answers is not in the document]
colder-is-better: Yes, I will update accordingly.

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
fghaas: I think you can use `curl` with a `telnet://` URL, no? I mean, you're not _actually_ doing HTTP here anywhere, so why pretend that you do?

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+```
+
+```plain
+Yello from srv-lbaas-2!
+```
+
+During our testing, executing the `wget` command repeatedly, we were
+getting those two hostnames one after the other --- and that was
+proof our TCP load balancer was working as expected. But there is
+one more expectation of any load balancer; the ability to skip
```
fghaas, line 586: I know I am being an extreme stickler here, but the semicolon is incorrect (since the clause coming after is has an object, but is lacking a subject and a predicate). :) 

```suggestion
one more expectation of any load balancer: the ability to skip
```

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+```bash
+wget -q 198.51.100.129:61234 -O -
+```
+
+```plain
+Yello from srv-lbaas-2!
+```
+
+We see that `srv-lbaas-2` responded. That means the next time we
+try to connect, `srv-lbaas-1` will respond. But if we SSH into
```
fghaas, line 602: Residual reference to SSH.

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+`srv-lbaas-1` and terminate `ncat`, then after connecting with
+`wget` we will get a response from `srv-lbaas-2` --- again. This,
+at least, is our expectation. So without further ado, we SSH into
```
fghaas, line 605: Another residual reference to SSH.

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+    you have to, and stop when the _Health Monitor_ section is
+    fully visible. You will see a message saying, "No health
+    monitor created", so click the green _Create a Healthmonitor_
+    button to create one.
+    
+    ![Create new health monitor](assets/shot-24.png)
+    
+    Pick a name for the new health monitor, and set its type to
+    PING. That way, ICMP echo requests will be used to determine
+    whether pool members are accessible.
```
fghaas, line 650: Let's please not do that; that's another bad practice we'd prefer users not to do. Whether a _host_ is pingable really has no bearing on whether the _service_ is accessible. If the host is up but the service has been stopped, we want that host to be dropped from the pool, and when we use a `PING` health monitor we don't. So let's use a `TCP` one instead.

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
colder-is-better: I just tried with `telnet://` and it doesn't work. So, as fae as I can tell, it's going to be either `wget -q p.q.r.xyz:61234 -O -` or `curl --http0.9 p.q.r.xyz:61234`.

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: Oops, I admit I am not aware of colon/semicolon-related rules :/

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
fghaas: I just followed the steps on my local machine and then ran `curl -s telnet://localhost:61234` and got the expected response back...

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+    ```
+
+## Testing the load balancer
+
+Each of the two test servers runs `ncat` to continuously listen
+on port 61234/TCP and respond to clients with a simple message,
+revealing its hostname:
+
+```bash
+ncat -kV -l 61234 -c 'echo Yello from $(hostname)!'
```
fghaas, line 555: I think this may have been accidentally changed in 4698e923cdb46a6c604cd9ffb706a8305b1742a8. `-kv` is working for me, `-kV` is not.

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: No, it is not, I am just describing a course of action.

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: Again, it is not :)

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: I see --- and I will update accordingly!

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: That was accidental (and nice catch, BTW).

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
colder-is-better: Interesting. From Mac OS X, it doesn't work (you get nothing back). From a Raspberry Pi with Debian 11 it *does* work, but you have to hit [CTRL+C] to get your command line back. Let us use something that works on both platforms, meaning either `wget` or `curl` (with no `telnet://`). What do you think?

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+Each of the two test servers runs `ncat` to continuously listen
+on port 61234/TCP and respond to clients with a simple message,
+revealing its hostname:
+
+```bash
+ncat -kv -l 61234 -c 'echo Yello from $(hostname)!'
+```
+
+Since the load balancer has a floating IP (`198.51.100.129`,
+in our example). and we know the port it listens to, we can
```
fghaas, line 559: Looks like a full-stop slipped in here instead of a comma; sorry for not catching this earlier.

```suggestion
in our example), and we know the port it listens to, we can
```

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
fghaas: OK, let's use `wget` then. Apologies for the noise. :)

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,742 @@
...
+    | 76ddfdc73 |           |            |                     |           |               |                  |        |
+    | ca2       |           |            |                     |           |               |                  |        |
+    | 04a32731- | srv-      | dfc7004673 | ACTIVE              | 10.15.25. |         61234 | ONLINE           |      1 |
+    | f3a6-     | lbaas-2   | 96428bacba |                     | 236       |               |                  |        |
+    | 41ca-     |           | 4376e72cc3 |                     |           |               |                  |        |
+    | a638-     |           | e9         |                     |           |               |                  |        |
+    | 6f8767437 |           |            |                     |           |               |                  |        |
+    | 50c       |           |            |                     |           |               |                  |        |
+    +-----------+-----------+------------+---------------------+-----------+---------------+------------------+--------+
+    ```
```
fghaas, line 742: There's one more note I would include here (because it is frequently a source of confusion), which is what `openstack loadbalancer member list` looks like after you kill `ncat` on one server, and what your `wget` command does under the same circumstances (namely, that it hits only the remaining healthy server, rather than alternately hitting a healthy and a broken one).

docs/howto/openstack/octavia/lbaas-tcp.md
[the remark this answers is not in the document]
colder-is-better: `wget` it is, then --- and no noise at all!

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: **I'm** sorry for not seeing that :/

docs/howto/openstack/octavia/lbaas-tcp.md
colder-is-better: Ah, yes, that note would be useful. Adding and updating in a bit...

fghaas: Ack, and in the process please also fix [the Markdown lint errors](https://github.com/citynetwork/docs/actions/runs/3986415488/jobs/6834966865#step:6:33) that the CI runs highlight.

colder-is-better: Yeah, although I run `tox -e markdownlint` before updating, sometimes I accidentally open the file from the *wrong* editor, errors creep in, and of course, I forget to re-run the linter :/

docs/howto/openstack/octavia/lbaas-tcp.md
```diff
@@ -0,0 +1,827 @@
...
+remote service via the load balancer:
+
+```bash
+wget -q 198.51.100.129:61234 -O -
+```
+
+```plain
+Yello from srv-lbaas-2!
+```
+
```
fghaas, line 821: I think you can drive the point home even better if you instruct the learner to run the command three times, like so: (note that the GitHub web interface chops off the closing code fence here; you'll have to re-instate it.)

```suggestion
Finally, let us check what happens when we try to access the
remote service via the load balancer, multiple times:

```console
$ wget -q 198.51.100.129:61234 -O -
Yello from srv-lbaas-2!

$ wget -q 198.51.100.129:61234 -O -
Yello from srv-lbaas-2!

$ wget -q 198.51.100.129:61234 -O -
Yello from srv-lbaas-2!
```

```

colder-is-better: Sure, I made some changes taking into account your suggestion!

fghaas (approved):
I think this is good to merge now. Any second thoughts, or reasons why this *shouldn't* be merged/deployed?

fghaas:
(Just rebased on current `main`, no other changes.)

colder-is-better:
I tried, but I cannot think of a reason not to merge/deploy :)

PR #117, opened 2023-01-17
feat: Add Howto section: Ansible

andersuno:
Section of Ansible howto's to provide value for customers in their automation journey.

### DIFF
--- a/docs/howto/ansible/index.md
+++ b/docs/howto/ansible/index.md
@@ -1,0 +1,5 @@
+# Ansible
+
+Ansible is a configuration management and automation tool. Get started by deploying an Ansible control node.
+
+[Official Ansible Docs](https://docs.ansible.com/)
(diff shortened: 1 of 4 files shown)


docs/howto/ansible/index.md
senare, line 1: Hope you don't mind If I inject myself and comment some ...

This is assuming push mode for ansible (which I guess is default starting point) but it would be nice to mention the ability to do pull also somewhere .. maybe using cloud init ?

For some explanation if you don't follow my trace of mind could be this (which I googled and haven't actually tried but to explain the concept I am referring to I hope its fine!) REF => https://jpmens.net/2012/07/14/ansible-pull-instead-of-push/

fghaas: Thanks for the comment. This PR won't go in in its current state and I forgot to close it. We'll do that now, and we'll look into covering Ansible at a later date.

fghaas:
We'll revisit this at a later date and will close this PR for now.

PR #121, opened 2023-01-31
refactor: Replace direct links to legal documents with macros

fghaas:
Since it might come in handy using references to legal documents
in multiple spots in the documentation, define macros for them.


colder-is-better (approved):
LGTM!

PR #123, opened 2023-02-08
refactor: Add macro placeholder for "Gardener"

fghaas:
Since the service will launch with some name that will most probably not be "Gardener", replace the hard-coded mentions thereof with a macro reference.


fghaas:
@colder-is-better I've rebased this on current `main` so I will need another approval, please. Thanks!

colder-is-better (approved):

PR #127, opened 2023-02-09
build: Remove PDF export plugin

fghaas:
We've had PDF export disabled in our builds for a while now (since #111), and nobody appears to be missing it.
    
Thus, remove the PDF export plugin, which also means we no longer need to install the `weasyprint` package in the build environment (which, in turn, speeds up the CI build checks).


colder-is-better (approved):
If there's a person missing that make-a-pee-dee-eff thing, I would like to talk to them (or maybe not).

PR #129, opened 2023-02-15
fix: Apply minor edit (syntax)

colder-is-better:
Delete extra word ("selecting")


fghaas (approved):
Eeeek. Thanks for spotting this. :)

colder-is-better:
I should have seen it earlier :/

PR #130, opened 2023-02-21
build: Replace generic issue templates with issue forms

fghaas:
Convert the remaining pure Markdown issue report templates to forms.

Also, add automatic issue labels.


.github/ISSUE_TEMPLATE/textual_bug.yml
```diff
@@ -0,0 +1,56 @@
...
+        Quote the passage that you believe needs improvement.
+      label: What the documentation says
+    id: actual
+    type: textarea
+    validations:
+      required: true
+  - attributes:
+      description: >
+        Make a suggestion on how to improve the passage and to make it
+        more understandable, more accessible, or simply more correct.
```
colder-is-better, line 38:
```suggestion
        more understandable, more accessible, or simply better.
```

.github/ISSUE_TEMPLATE/functional_bug.yml
```diff
@@ -0,0 +1,89 @@
...
+        it here. Otherwise, you can leave this blank.
+      label: Additional context
+    id: context
+    type: textarea
+    validations:
+      required: false
+description: >
+  Use this template to report functionally incorrect or misleading
+  documentation.
+labels: functional-bug
```
colder-is-better, line 88:
```suggestion
label: functional-bug
```

.github/ISSUE_TEMPLATE/gap.yml
```diff
@@ -57,4 +57,5 @@ body:
 description: >
   Use this template to report documentation gaps: information that is
   missing altogether, or that is incomplete.
+labels: gap
```
colder-is-better, line 60:
```suggestion
label: gap
```

.github/ISSUE_TEMPLATE/textual_bug.yml
```diff
@@ -0,0 +1,56 @@
...
+        If there’s any other context you’d like to share, please add
+        it here. Otherwise, you can leave this blank.
+      label: Additional context
+    id: context
+    type: textarea
+    validations:
+      required: false
+description: >
+  Use this template to report grammar glitches, syntax errors, and typos.
+labels: textual-bug
```
colder-is-better, line 55:
```suggestion
label: textual-bug
```

.github/ISSUE_TEMPLATE/functional_bug.yml
fghaas: What, you think we get to make up our own syntax? :)

I'm afraid we'll have to follow [what GitHub gives us](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax).

.github/ISSUE_TEMPLATE/textual_bug.yml
fghaas: [Nope](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax).

.github/ISSUE_TEMPLATE/textual_bug.yml
fghaas: No, I consciously used "more correct" here.

.github/ISSUE_TEMPLATE/gap.yml
fghaas: Again, [nope](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#top-level-syntax), sorry. :)

.github/ISSUE_TEMPLATE/functional_bug.yml
```diff
@@ -0,0 +1,89 @@
+# yamllint disable rule:line-length
+---
+body:
+  - attributes:
+      description: >-
```
colder-is-better, line 5: I see that most of the time you use `>` for multiline strings, but there are some times you use `>-` (which means "don't append a linebreak at the end", if I'm not mistaken). Is there a reason for that?

colder-is-better (commented):
All is looking good -- I only have some pico-nits and a question.

.github/ISSUE_TEMPLATE/functional_bug.yml
fghaas: Not really. Emacs just automatically turns `>` into `>-`.

But GitHub's issue forms renderer doesn't even distinguish between `>` (multiline string, drop newlines) and `|` (multiline string, retain newlines); it always behaves as if using `|`.

colder-is-better (approved):

PR #138, opened 2023-02-23
fix: Correct link to SSE-C from Compliant Cloud features reference

fghaas:
Fix the broken link to the S3 SSE-C howto from the Compliant Cloud feature reference page.

### DIFF
--- a/docs/reference/features/compliant.md
+++ b/docs/reference/features/compliant.md
@@ -28,11 +28,11 @@
 
 ## Object storage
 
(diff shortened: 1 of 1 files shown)


fghaas:
@colder-is-better I don't really know how this slipped past htmlproofer link checking. We can look into that separately, but we'll want to fix the link before doing anything else. So I'd appreciate a quick review on this one.

The Public Cloud feature list does not have this problem.

colder-is-better (approved):

PR #142, opened 2023-03-03
fix: Mention "openstack server ssh"

fghaas:
In our "new server" how-to guide, we previously omitted the ability to connect to the new instance via SSH, using the `openstack server ssh` command.

Fix that omission.


colder-is-better (approved):
I have forgotten about that Zug!

PR #144, opened 2023-03-09
fix: Drop CCMP references from Magnum cluster how-to guide

fghaas:
CCMP-launched Magnum clusters currently do *not* allow OpenStack CLI users to successfully download cluster credentials.
    
This, sadly, renders CCMP-launched Magnum clusters functionally unusable for use with Kubernetes: one can spin up a cluster, but one has no way of pointing `kubectl` at it and actually do anything with the cluster.
    
Thus, until a kubeconfig download button is implemented in CCMP, remove the CCMP-related bits from the Magnum cluster spin-up howto, and declare that functionality CLI-only.

This PR also includes a number of smaller changes and chores, including a reformat of the guide.

But the important changes are in 2e55ed1009f80465dee08c2fd4160aae8c4f26d8 — I have put them in a separate commit so they should be straightforward to revert, once the missing functionality has been implemented in CCMP.

### DIFF
--- a/docs/howto/openstack/magnum/new-k8s-cluster.md
+++ b/docs/howto/openstack/magnum/new-k8s-cluster.md
@@ -1,28 +1,21 @@
 # Creating new Kubernetes clusters
 
-By employing OpenStack
-[Magnum](https://docs.openstack.org/magnum/latest) you can easily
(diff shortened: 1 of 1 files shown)


colder-is-better (approved):
Looking good!

PR #150, opened 2023-03-15
fix: Update Ceph service version for Sto2 and Tky1

revij:
We now run Ceph Quincy in Sto2 and Swift/S3 are available in Tky1.


fghaas (commented):
This looks great but please include a corresponding change to `docs/reference/features/public.md` as well (you'll need to change a few instances of `:material-close:` to `:material-check:` for Tky1).

Thanks!

fghaas (approved):
Took care of the remaining change. This is good to merge.

PR #151, opened 2023-03-16
feat: Add HowTo on hibernating/waking K8s clusters

colder-is-better:
From the Cleura Cloud Management Panel, we show how we can put Gardener-based Kubernetes clusters in hibernation, and how we can wake up already hibernated clusters.

PR #153, opened 2023-03-20
Add information about dynamically-managed DNS records in Gardener clusters

fghaas:
* Split the Gardener "create shoot cluster" how-to into two guides, by factoring out the information on getting a kubeconfig into a separate page.
* Explain how Gardener manages the DNS entry for the API server's address.

Since this makes a cross-reference to hibernation, this PR can land only after #151 does.

### DIFF
--- a/docs/howto/kubernetes/gardener/kubectl.md
+++ b/docs/howto/kubernetes/gardener/kubectl.md
@@ -15,6 +15,8 @@ Create the directory and the file if needed.
 
 > By default, `kubectl` searches for its configuration in `~/.kube/config`, but you can [modify this behavior](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) by setting the `KUBECONFIG` environment variable, if needed.
 
+## Verifying your kubeconfig
+
 Check if your `kubectl` uses the proper configuration by running:
 
 ```shell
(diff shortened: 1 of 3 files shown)


docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -0,0 +1,95 @@
+# Managing a Kubernetes cluster
+
+## Extracting the kubeconfig file
+
+Once you [have configured a new shoot cluster](create-shoot-cluster), you need to create a [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) file to be able to access it.
```
colder-is-better, line 5:
```suggestion
Once you [have instantiated a new shoot cluster](create-shoot-cluster), you need to create a [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) file to be able to access it.
```

Since we are configuring *and* creating a new shoot cluster, I suggest a slight change in wording, like the above.

colder-is-better (commented):
Looking good! (I have one pico-suggestion only.)

docs/howto/kubernetes/gardener/kubectl.md
fghaas: I understand why you're not suggesting "created" here (it produces a repetition), but "instantiated" violates Orwell's [Rule № 5](https://en.wikipedia.org/wiki/Politics_and_the_English_Language#Remedy_of_Six_Rules), so I went with "launched".

colder-is-better: Well, my issue with "configured" was that it's not enough, because you actually configure and **then** create a new shoot cluster. So because that thought stuck in muh mind and wouldn't go away, I came up with "instantiated" and immediately started breathing better :)

colder-is-better (approved):

PR #158, opened 2023-03-27
feat: Add BG info on Gardener auto-upgrading

colder-is-better:
Explain how auto-upgrading works, especially for clusters which during the maintenance window are hibernating.


murwall:
LGTM

fghaas:
@colder-is-better Please let me know if you have any objections, otherwise I'd go ahead and merge this.

docs/background/kubernetes/gardener/garden-linux.md
```diff
@@ -0,0 +1,29 @@
+---
+description: Gardener in Cleura Cloud runs its worker and control plane nodes on Garden Linux.
+---
+# Garden Linux
+
+In a {{k8s_management_service}} cluster, worker nodes run [Garden Linux](https://gardenlinux.io), which is a [Linux distribution](https://en.wikipedia.org/wiki/Linux_distribution) based on [Debian GNU/Linux](https://debian.org/).
+
+Garden Linux provides an automated, auditable build pipeline.
+Its snapshot [releases](https://github.com/gardenlinux/gardenlinux/releases) are automatically generated from GitHub Actions workflows.
```
colder-is-better, line 9: It's probably just me, but I have to wonder: Why are we providing these pieces of info regarding automated, auditable build pipelines, and snapshots generated from GitHub Actions workflows?

colder-is-better (commented):
Looks ready to go (I have one question only).

docs/background/kubernetes/gardener/garden-linux.md
colder-is-better: I am not saying otherwise. All I'm saying is that this piece of info came in somewhat unexpectedly. At least this is how I felt, as a reader.

fghaas: Fair. Then it's probably unneeded. I've dropped that paragraph.

fghaas (approved):

PR #161, opened 2023-03-29
fix: Use macro instead of "Gardener"

fghaas:
In several places, the use of the string literal "Gardener" has slipped in where we should be using the `k8s_management_service` macro.

Replace the literal with the macro where applicable (this excludes the YAML front matter, and first-level headings).


colder-is-better (approved):
All replacements look good!

PR #164, opened 2023-04-03
refactor: Split kubeconfig information for Magnum into a separate page

fghaas:
* Create a new page, `kubectl.md`, in the OpenStack Magnum how-to guides section (similar to what we already have for Gardener).
* Explain that one must use CCMP to download the kubeconfig of a CCMP-managed Magnum cluster, and the CLI for a CLI-managed one.
* Instead of specifically discussing how to install kubectl on Debian/Ubuntu and MacOS, point to the upstream documentation.

### DIFF
--- a/docs/howto/openstack/magnum/kubectl.md
+++ b/docs/howto/openstack/magnum/kubectl.md
@@ -1,0 +1,91 @@
+---
+description: How to fetch, verify, and use your kubeconfig with kubectl in a Magnum-managed Kubernetes cluster.
+---
+# Managing a Kubernetes cluster
+
+Once you [have launched a new cluster](new-k8s-cluster.md), you can interact with it using `kubectl` and a [kubeconfig](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) file.
+
+## Prerequisites
+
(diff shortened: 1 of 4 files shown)


docs/howto/openstack/magnum/kubectl.md
```diff
@@ -0,0 +1,91 @@
...
+    kubeconfig--<cluster_name>--<region_name>--<alphanum_id>.yaml
+    ```
+
+    Feel free to rename it to something simpler, like `config`.
+=== "OpenStack CLI"
+    To download the kubeconfig file for your Kubernetes cluster, type
+    the following:
+
+    ```bash
+    openstack coe cluster config --dir=${PWD} bangor
```
colder-is-better, line 40: Since Bangor is not on our radar in this HOWTO, I suggest this change:

```suggestion
    openstack coe cluster config --dir=${PWD} <cluster_name>
```

fghaas: Ah yes, absolutely. Will fix.

colder-is-better (commented):
It's *almost* ready to go!

colder-is-better (approved):

PR #165, opened 2023-04-04
fix: Add screenshot with new machine img indicator

colder-is-better:
When there is a new machine image for a Gardener cluster, before the automatic upgrade kicks in during the maintenance window, there is a relative indication in the detailed view of the cluster. Add a screenshot that clearly shows what an indication like this looks like.


docs/background/kubernetes/gardener/assets/gard_new_img_avail.png
fghaas, line 1: May want to obfuscate the `pNNNNN` bits here.

docs/background/kubernetes/gardener/autoupgrades.md
```diff
@@ -15,6 +15,13 @@ During the upgrade, the API server may briefly be unavailable.
 Both minor and major [Garden Linux](garden-linux.md) release upgrades are allowed and performed automatically.
 Also, version jumps are allowed over minor release upgrades.
 
+Whenever there is a new machine image available, this is indicated in
+the {{gui}}. Before the auto-upgrade kicks in, the version of the new
+image is displayed in the expanded view of the cluster, in the *Worker
+Groups* tab.
```
fghaas, line 21: A few notes here:

(1) Since this Markdown file now uses the "1 sentence per line" convention, please keep it that way. Thus, please make this:

```suggestion
Whenever there is a new machine image available, this is indicated in the {{gui}}.
Before the auto-upgrade kicks in, the version of the new image is displayed in the expanded view of the cluster, in the *Worker Groups* tab.
```

(2) I understand that upgradeable nodes are also marked with a small arrow in the overview, maybe we should show that here as well?

(3) Perhaps we should even *just* include the screenshot with the arrow markers in *this* section of the docs, and include the one originally included here in a separate how-to on manual upgrades?

colder-is-better: (1) One sentence per line it is!

(2) I had to recreate the cluster to make sure: When there's a new machine image available, there ain't no marks/small arrows for the nodes --- not in the overview, not anywhere.

(3) Those arrow markers? Nowhere to be seen :/

docs/background/kubernetes/gardener/autoupgrades.md
```diff
@@ -15,6 +15,14 @@ During the upgrade, the API server may briefly be unavailable.
 Both minor and major [Garden Linux](garden-linux.md) release upgrades are allowed and performed automatically.
 Also, version jumps are allowed over minor release upgrades.
 
+Whenever there is a new machine image available, this is indicated in the {{gui}} --- and in the cluster row in particular.
```
fghaas, line 18: The clause after the em dash sounds odd. I'd suggest to drop it, and just replace it with a colon:

```suggestion
Whenever there is a new machine image available, this is indicated in the {{gui}}:
```

fghaas (commented):
Just one more tiny nit, and then I think this is good to go. :)

fghaas (approved):
Looks good now, thanks!

PR #170, opened 2023-04-06
fix: Be more specific re. autoupgrades in Gardener

colder-is-better:
When we create a Kubernetes cluster with Gardener, there is a maintenance window during which automatic upgrades take place. This feature is specific to Cleura Cloud, so it makes sense to clarify that.

### DIFF
--- a/docs/background/kubernetes/gardener/autoupgrades.md
+++ b/docs/background/kubernetes/gardener/autoupgrades.md
@@ -3,7 +3,7 @@ description: How auto-upgrading works in Gardener-managed Kubernetes clusters
 ---
 # Automatic upgrades
 
-By default, Kubernetes clusters created with {{k8s_management_service}} are upgraded automatically.
+By default, Kubernetes clusters created with {{k8s_management_service}} in {{brand}} are upgraded automatically.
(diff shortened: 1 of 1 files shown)


docs/background/kubernetes/gardener/autoupgrades.md
```diff
@@ -3,7 +3,7 @@ description: How auto-upgrading works in Gardener-managed Kubernetes clusters
 ---
 # Automatic upgrades
 
-By default, Kubernetes clusters created with {{k8s_management_service}} are upgraded automatically.
+By default, Kubernetes clusters created in {{brand}} with {{k8s_management_service}} are upgraded automatically.
```
fghaas, line 6: The product name is "Gardener in Cleura Cloud", hence:

```suggestion
By default, Kubernetes clusters created with {{k8s_management_service}} in {{brand}} are upgraded automatically.
```

fghaas (commented):
Just one nit, and when you force-push please make sure you're rebased on current `main`.

fghaas (approved):
Looks good to me now, thanks!

PR #171, opened 2023-04-11
feat: Add limitations for nested virtualization

fghaas:
Explain that we only support Linux guests for nested virtualizations, and that they must run reasonably recent kernel and KVM versions and pass in the pcid CPU feature flag.

### DIFF
--- a/docs/reference/limitations/openstack.md
+++ b/docs/reference/limitations/openstack.md
@@ -20,6 +20,17 @@ It also means that you cannot use the following `openstack` CLI commands; they a
   > You can use `openstack project list`, but this will only list the project(s) that your user account has access to.
 * `openstack domain <subcommand>`
 
+## Nova
+
+### Nested virtualization
+
+Running nested virtualization is only supported for Nova instances ([servers](../../howto/openstack/nova/new-server.md)) running Linux.
+The server must run a [Linux kernel](https://www.kernel.org/) of version 5.0 or later, and [QEMU/KVM](https://www.qemu.org/) 4.1 or later.
+
+This means that you can run nested virtualization on servers booted from a CentOS 9 (or later), Ubuntu 20.04 (or later), or Debian 11 (or later) base image.
+
+Furthermore, you must ensure that the Nova server [passes the `pcid` CPU feature flag](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#important-cpu-features-for-intel-x86-hosts) to nested guests.
+
 ## Neutron
 
 ### Dynamic routing


noonedeadpunk (dismissed):
LGTM

docs/reference/limitations/openstack.md
```diff
@@ -20,6 +20,15 @@ It also means that you cannot use the following `openstack` CLI commands; they a
   > You can use `openstack project list`, but this will only list the project(s) that your user account has access to.
 * `openstack domain <subcommand>`
 
+## Nova
+
+### Nested virtualization
+
+Running nested virtualization is only supported for Nova instances running Linux as the guest operating system.
```
colder-is-better, line 27: Nova instances act like hosts, right? If so, then I would suggest this change:
```suggestion
Nested virtualization only works in Nova instances running Linux.
```

docs/reference/limitations/openstack.md
```diff
@@ -20,6 +20,15 @@ It also means that you cannot use the following `openstack` CLI commands; they a
...
+The guest must run a [Linux kernel](https://www.kernel.org/) of version 5.0 or later, and [QEMU/KVM](https://www.qemu.org/) 4.1 or later.
```
colder-is-better, line 28: Also, if I understand correctly, I believe this is more accurate:
 ```suggestion
Guests must be created with [QEMU/KVM](https://www.qemu.org/) 4.1 or later, and should run a [Linux kernel](https://www.kernel.org) of version 5.0 or later.
```

File: docs/reference/limitations/openstack.md [hunk 3]
```diff
@@ -20,6 +20,15 @@ It also means that you cannot use the following `openstack` CLI commands; they a
...
+
+Furthermore, you must ensure that the Nova instance [passes the `pcid` CPU feature flag](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#important-cpu-features-for-intel-x86-hosts) to the nested guest.
```
colder-is-better, line 30:
```suggestion
Furthermore, you must ensure that the Nova instance [passes the `pcid` CPU feature flag](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#important-cpu-features-for-intel-x86-hosts) to nested guests.
```

colder-is-better (commented):
Looking good, I have a question and a couple of suggestions only.

docs/reference/limitations/openstack.md
fghaas: Yes, can do.

docs/reference/limitations/openstack.md
fghaas: Done.

docs/reference/limitations/openstack.md
fghaas: I rephrased this slightly, and added information on what base images should be good to use for this purpose.

colder-is-better: Very nice!

colder-is-better (approved):

PR #176, opened 2023-04-13
fix: Explain & show how to get an RC file directly

colder-is-better:
While trying to get the RC file for an OpenStack user, there is now no option for generating said file but an option for downloading it directly. Modify the text and update the screenshot in the HOWTO, to reflect that change.


colder-is-better:
The screenshots have *not* been updated yet.

fghaas:
> The screenshots have _not_ been updated yet.

Right, in that case the PR should be marked as Draft, so that we don't merge it unintentionally. :)

fghaas (commented):
This looks good to me (modulo the missing screenshot update, of course).

PR #178, opened 2023-04-14
fix: Make sure variable name is properly referenced

colder-is-better:
Encase legal_docs.transfer_form.name in double curly brackets, so the variable value is used in the provided URL (legal_docs.transfer_form.url).

While we're at it, apply the one-sentence-per-line policy and introduce some minor changes to syntax and prose.

### DIFF
--- a/docs/howto/account-billing/change-billing-data.md
+++ b/docs/howto/account-billing/change-billing-data.md
@@ -1,29 +1,21 @@
 # Changing your account billing data
 
-Via [{{gui}}](https://{{gui_domain}}) you can change the contact
-person, address, company name, and purchase order number associated
-with your account.
+You may at any time change the contact person, address, company name, and purchase order number associated with your {{brand}} account.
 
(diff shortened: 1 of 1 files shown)


fghaas:
> Encase legal_docs.transfer_form.name in double curly brackets, so the variable
value is used in the provided URL (legal_docs.transfer_form.url).

Ooof. Sorry about that.

> While we're at it, set the maximum text width to 72 and introduce some minor
changes to syntax and prose.

If we're already making a broad change, then let's change to one-sentence-per-line instead.

colder-is-better:
I wonder why the build fails:

```
ERROR - invalid url - https://s3-kna1.citycloud.com:8080/6a5aa55d8f094a13ae18199639aa72c2:cleura.files/Cleura-Form-Termination-Of-Subscription.pdf [-1] [howto/account-billing/delete-account.md]
```

The URL seems to be working fine :/

fghaas (approved):
Looks good to me! I'll rekick the build checks.

PR #180, opened 2023-04-14
fix: Drop :8080 port suffix from public S3 links

fghaas:
Our object storage services now support access via the standard HTTPS port, 443. This transition has not yet been completed in all regions, so we cannot update the S3 and Swift API how-to guides yet, but it has been completed in the Kna1 region so we can update the references to the publicly accessible legal documents.

Thus, drop the `:8080` port suffix from those macro definitions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,9 @@ jobs:
       - name: Install Python dependencies
         run: |
           pip install tox
-      - name: Test with tox
+      - env:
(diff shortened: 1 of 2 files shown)


colder-is-better (dismissed):
LGTM!

fghaas:
Trying to unbreak the build with [b87ec4e](https://github.com/citynetwork/docs/pull/180/commits/b87ec4e9620ac6e8549a0d619d34fed1920aad2d).

colder-is-better (approved):

PR #181, opened 2023-04-19
fix: Use shortcut for account registration

fghaas:
Rather than sending customers to the CCMP start page and then walking them to the account creation page, we can instead instruct them do drop directly on the registration page.

This saves one step, and removes the need for one screenshot.


colder-is-better (approved):
LGTM!

PR #184, opened 2023-04-21
feat: Expand account billing data HOWTO with e-invoicing

colder-is-better:
If certain conditions are met, a customer may request e-invoicing. We add a new section in the existing HOWTO, listing the requirements and describing the process of requesting e-invoicing.


docs/howto/account-billing/change-billing-data.md
```diff
@@ -19,3 +19,30 @@ Finalize your changes by clicking the green _Update_ button at the bottom.
...
 In case you do wish to change any of those, you will have to submit a [{{legal_docs.transfer_form.name}}]({{legal_docs.transfer_form.url}}) via our [{{support}}](https://{{support_domain}}/servicedesk).
+
+## E-invoicing for Swedish customers
+
+This is available to select accounts.
+More specifically, you have the option of e-invoicing if all three conditions are met:
+
+* the country in the customer contract is set to _Sweden_ (that is the country in the account owner data, not the country in the billing data),
+* you have provided a company name,
+* you have provided a Swedish [organisationsnummer](https://www.bolagsverket.se/foretag/organisationsnummer.1207.html) (link available in Swedish only).
```
fghaas, line 30: I would put the non-English term in italics, and shorten the parenthetical expression:
```suggestion
* you have provided a Swedish *[organisationsnummer](https://www.bolagsverket.se/foretag/organisationsnummer.1207.html)* (link in Swedish).
```

docs/howto/account-billing/change-billing-data.md
```diff
@@ -19,3 +19,31 @@ Finalize your changes by clicking the green _Update_ button at the bottom.
 Please note that you cannot readily change your organization number (for business accounts) or your personal number (for individual accounts) in the {{gui}}.
 
 In case you do wish to change any of those, you will have to submit a [{{legal_docs.transfer_form.name}}]({{legal_docs.transfer_form.url}}) via our [{{support}}](https://{{support_domain}}/servicedesk).
+
+## E-invoicing for Swedish customers
```
fghaas, line 23: I think we can simplify this to just:

```suggestion
## E-invoicing
```

fghaas (commented):
I think this looks great (I have just a couple of nits), but if I look at the rendered documentation I come to the conclusion that this is really better suited for a separate page (called, say `e-invoice.md`), which would also facilitate linking to it in customer communications.

What do you think?

andersc2022:
> 

I agree Florian - This documentaion is better suited for a separate page when we communicates the link to our customers.

fghaas:
> I agree Florian - This documentaion is better suited for a separate page when we communicates the link to our customers.

Yes, though let it be known it was I who suggested putting it into the account & billing how-to in the first place. Apologies to @colder-is-better for the change of heart.

colder-is-better:
The original HOWTO is about changing customer billing data, and e-invoicing has to do with billing data. In my mind, there is no need for a separate HOWTO. If we created a separate HOWTO, it would be a weirdly short document applicable only to a part of our customer base. (I would even say that, to some, it would look a bit discriminatory, and then you would have many non-Swedish customers requesting similar functionality for them.) In any case, I believe the original HOWTO is the place to be for what this really is: just a section. (By the way, linking directly to a section is still easy.)

fghaas:
> The original HOWTO is about changing customer billing data, and e-invoicing has to do with billing data.

Exactly, but customers may also want to enable e-invoicing right when they create an account. Thus, I'd argue that it's helpful to have a separate page that covers e-invoices, which we can then cross-reference from both the account creation and the account & billing data how-to guides.

PR #185, opened 2023-04-21
fix: Update Ceph version in Sto1HS region

colder-is-better:
Ceph Quincy is now deployed and available in Sto1HS region, in the compliant cloud.


fghaas:
Rebased on current `main`.

fghaas (approved):

PR #186, opened 2023-04-26
feat: Mention default storage class in Magnum how-to

fghaas:
Virtually all sensible usage scenarios for Kubernetes require the use of a storage class for dynamic volume provisioning, and Magnum does not define any such thing.

Thus, add instructions to define a default storage class.

### DIFF
--- a/docs/howto/openstack/magnum/kubectl.md
+++ b/docs/howto/openstack/magnum/kubectl.md
@@ -89,3 +89,31 @@ kube-system   magnum-metrics-server-79556d6999-xdlpm       1/1     Running   0  
 kube-system   npd-5p6gk                                    1/1     Running   0          165m
 kube-system   openstack-cloud-controller-manager-44rz9     1/1     Running   0          167m
 ```
+
+## Defining a default storage class
+
+An OpenStack Magnum-managed cluster does not automatically define a default [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/) for [dynamic volume provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/).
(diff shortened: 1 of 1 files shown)


docs/howto/openstack/magnum/kubectl.md
```diff
@@ -89,3 +89,31 @@ kube-system   magnum-metrics-server-79556d6999-xdlpm       1/1     Running   0
 kube-system   npd-5p6gk                                    1/1     Running   0          165m
 kube-system   openstack-cloud-controller-manager-44rz9     1/1     Running   0          167m
 ```
+
+## Defining a default storage class
+
+An OpenStack Magnum-managed cluster does not automatically define a default [storage class](https://kubernetes.io/docs/concepts/storage/storage-classes/) for [dynamic volume provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/).
+You should define it immediately upon cluster creation.
```
colder-is-better, line 96: Following on the previous sentence, I would change that one into the following:

```suggestion
So after creating your new cluster, you should define a storage class immediately.
```

(The way I see it ---and I may be wrong on the percentage---, leaving the original sentence would have 1,27% of the readers believe that the whole section was generated by an AI.)

colder-is-better (commented):
99,7649% there!

docs/howto/openstack/magnum/kubectl.md
fghaas: Erm, no. That doesn't sound like an improvement to me. This does, tough:
```suggestion
You should define one immediately upon cluster creation.
```

colder-is-better (approved):

PR #187, opened 2023-04-27
feat: Add reference information on object storage quotas

fghaas:
* Explain that the number of objects per bucket is limited.
* Add a .pages file so we get shorter headings in the navigation.
* Fix case in the title of the "OpenStack quotas" section.

Also, correct the path in our support URL (in several spots).


pejotes:
LGTM

PR #188, opened 2023-05-02
fix: Remove references to the old General Terms and Conditions

fghaas:
The transition period, during which the old GT&C still applied to accounts created before 2023-04-05, ends on 2023-05-05. As of that date, the new ToS applies to all accounts.

Thus, remove the remaining references to the old terms.


colder-is-better (approved):
LGTM!

armanborghem (approved):

PR #190, opened 2023-05-03
feat: Add background information on Disaster Recovery

colder-is-better:
We introduce the Disaster Recovery feature, explain what kind of Cleura objects can use it, and point out why it makes sense to keep it enabled. While we are at it, we add a new HOWTO guide on how to restore a server to a particular snapshot.


docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,54 @@
+# Disaster recovery
+
+When you go on to create a new server in {{brand}}, you will probably
```
fghaas, line 3: Not sure what "go on to" means here. We might be able to drop it (as we might drop "probably")?

```suggestion
When you create a new server in {{brand}}, you will
```

Also, please make "create a new server" a link to the corresponding how-to guide.

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,54 @@
...
+
+![The disaster recovery feature can be manually activated for any
+server](assets/disaster-recovery-option-manual-activation.png)
+
+In the following, we explain what this option does, how it works in the
+background, and why you should consider enabling it.
+
+## What it is
+
+The *Disaster Recovery* feature, DR for short, is managed from the
```
fghaas, line 21:
```suggestion
The *Disaster Recovery* (DR) feature is available via the
```

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,54 @@
...
+{{gui}} only and applies to servers and volumes that use our
+[Ceph](https://ceph.io) backend.
```
fghaas, line 23: The only way for this to _not_ be the case would be if a server uses an `s` flavor (see https://docs.cleura.cloud/reference/flavors/#compute-tiers), so let's explain that here. Any server that uses another flavor, and any volume, does use Ceph.

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,54 @@
...
+then all snapshots will also be deleted, and the snapshot creation
+schedule will be canceled immediately.
+
+## Why enable it
+
+Provided snapshots are available, you can restore a server or a single
+volume to any of those snapshots. For instance, you may discover that
+due to faulty application logic or simply a bug, you are now
+experiencing data corruption. Then, one of your options would be to go
+back in time by restoring one of the available snapshots and keep going
```
fghaas, line 43: This paragraph should contain a link to a short how-to guide for resetting a volume or server to the most recent snapshot state.

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,54 @@
...
+from there.
+
+You should know that the DR feature creates *point-in-time* snapshots on
+the storage level. This is highly advantageous because the process is
+almost instantaneous, with no write-lock on the RBD image. On the other
+hand, the time required to restore a particular image depends on its
+size. During restoration, the volume is automatically detached before
```
fghaas, line 50: You'll need to explain how this works for a boot-from-volume server (which cannot detach its root volume).

fghaas (commented):
This is off to a good start; I have a few inline comments.

docs/background/disaster-recovery.md
colder-is-better: Done and done!

docs/background/disaster-recovery.md
colder-is-better: Got it!

docs/background/disaster-recovery.md
colder-is-better: Understood. I provided the above piece of info based on the attached Wiki ("Disaster recovery questions"). Since I do not know what happens behind the scenes with boot-from-volume servers, I will either need some pointers or we will have to ask for an expert's insight here.

docs/background/disaster-recovery.md
fghaas: I'd say we want to roll that into this same PR. Information on backups/snapshots is mildly useless unless accompanied by information on how to restore. :)

colder-is-better: ...aaaaand I rolled that in :)

docs/background/disaster-recovery.md
colder-is-better: @fghaas After reading @Jeeaaasus's explanation and trying to restore a boot-from-volume server to a particular snapshot, I realized that no detachment takes place. Instead, the instance shuts off, the restoration process completes, and then you have to manually power it back on. I am updating the original paragraph accordingly.

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,85 @@
...
+## What it is
+
+The *Disaster Recovery* (DR) feature, is available via the {{gui}} and
+applies to servers and volumes that use our [Ceph](https://ceph.io)
+backend. That would be **all** servers but the ones of the `s`
+[flavor](../../reference/flavors/#compute-tiers).
+
+## How it works
+
+As soon as you enable DR for a server or a single volume, then you start
```
fghaas, line 28: "As soon as" followed by "then" sounds a little awkward to me.

```suggestion
As soon as you enable DR for a server or a single volume, you start
```

docs/background/disaster-recovery.md
fghaas: I meant roll it into the same *PR*, not roll it into the same file. We want a *separate* how-to guide, *as part of this PR,* that cross-references the background info (and vice versa). The step-by-step guide for restoring doesn't belong in background information.

fghaas (commented):
Please see the inline comments. :)

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,85 @@
...
+
+![The disaster recovery feature can be manually activated for any
+server](assets/disaster-recovery-option-manual-activation.png)
+
+In the following, we explain what this option does, how it works in the
+background, and why you should consider enabling it.
+
+## What it is
+
+The *Disaster Recovery* (DR) feature, is available via the {{gui}} and
```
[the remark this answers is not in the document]
colder-is-better: Alert and alarm and sirens going crazy (and thanks for spotting that).

docs/background/disaster-recovery.md
colder-is-better: It is indeed :$

docs/background/disaster-recovery.md
colder-is-better: Ahem, rollin' in and updating --- but properly, this time aroond!

colder-is-better:
> I think this is in excellent shape now. One style note: although the term "restoration" is perfectly correct, I think to call the opposite of a backup or snapshot a "restore" is more common. What do you think?

It is more common, there's no doubt about it. It's just that, as a personal preference, when I can choose between a noun and a verb, I always go with the noun. But yes, let's go with "restore". If anything, it sounds more "natural" than the noun :)

docs/background/disaster-recovery.md
```diff
@@ -0,0 +1,56 @@
+# Disaster recovery
```
fghaas, line 1: Please add YAML front matter with a `description` field (like you already did in the howto).

fghaas (commented):
Perfect. Just one more nit and then I think we can merge this!

docs/background/disaster-recovery.md
colder-is-better: Ouch (and done).

fghaas (approved):

PR #192, opened 2023-05-19
feat: Add tutorial on managing cloud objects with Ansible

colder-is-better:
We introduce a tutorial on Ansible, guiding the reader through creating and managing various Cleura Cloud objects like keypairs, networks, and servers. We also instruct the user to update a server, automatically reboot if necessary, and install a Systemd service. Finally, the user is instructed to delete Cleura Cloud objects --- always with the help of Ansible.

### DIFF
--- a/docs/tutorials/.pages
+++ b/docs/tutorials/.pages
@@ -1,0 +1,4 @@
+title: "Tutorials"
+nav:
+  - index.md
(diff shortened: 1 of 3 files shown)


colder-is-better:
Ahem, this should go to The Academy (TM) instead :/

PR #179, opened 2023-04-14
fix: Update Ceph version name in Kna1 region

colder-is-better:
Ceph Quincy is now deployed and available in all public cloud regions, including Kna1.

PR #193, opened 2023-05-21
fix: Update Ceph version in Sto2HS region

colder-is-better:
Ceph Quincy is now deployed and available
in Sto2HS region, in the compliant cloud.


colder-is-better:
> Here now, the "fix" that I erroneously added to #179, and which I had to revert in #183, actually applies. :)

It sure does :)

fghaas (approved):

PR #198, opened 2023-06-29
feat: Document image usage in our regions

noonedeadpunk:
With introduction of automatic image rotation we should describe
new naming convention that users may rely on when searching or
reffering to some image in their automation toolings.


fghaas:
This is good information, but a substantial portion of it doesn't belong into a reference page but into a Howto instead. Is it OK with you if we grab this one and take it from here?

noonedeadpunk:
Yes, totally, feel free to, was just trying to help out a bit. Also had doubts if it's reeference or not, but wasn't able to come up where to split these 2.

fghaas:
Closing this one, it's been superseded by #208.

PR #199, opened 2023-06-29
chore: Add cache and build results to .gitignore

noonedeadpunk:
When running local tests, all build results appear in git diff, which
makes hard to distinguish real changes from produced by build
content. So it's quite trivial to accidentally revert intended changes
together with invalidating build/cache content.


noonedeadpunk:
How in the world I ended up with 3 commits here, when made only 1... I can't be more annoyed with the github flow then I am right now...

fghaas:
> How in the world I ended up with 3 commits here, when made only 1... I can't be more annoyed with the github flow then I am right now...

Doesn't look like this has much to do with GitHub; your change just wasn't based on `main`. Fixed now.

fghaas:
> When running local tests, all build results appear in git diff,

Uh, no they don't? `git diff` only shows you diffs for things that are already in the working tree.

```
$ tox -e build
build: commands[0]> mkdocs build --strict
INFO     -  [macros] - Macros arguments: {'module_name': 'main', 'modules': [], 'include_dir': '', 'include_yaml': [], 'j2_block_start_string': '', 'j2_block_end_string': '',
            'j2_variable_start_string': '', 'j2_variable_end_string': '', 'on_undefined': 'strict', 'on_error_fail': True, 'verbose': False}
INFO     -  [macros] - Extra variables (config file): ['analytics', 'brand', 'brand_domain', 'company', 'company_domain', 'gui', 'gui_domain', 'k8s_management_service',
            'legal_docs', 'rest_api', 'rest_api_domain', 'support', 'support_domain', 'support_email']
INFO     -  [macros] - Extra filters (module): ['pretty']
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /home/florian/git/cleura/docs/site
INFO     -  Documentation built in 13.37 seconds
  build: OK (14.10=setup[0.04]+cmd[14.06] seconds)
  congratulations :) (14.16 seconds)
$ git diff
```

... comes up with an empty `git diff` result.  Did you mean `git status`?

And not having `.tox`, `.cache`, and `site` in `.gitignore` is quite intentional, since we want you to be able to return to a pristine checkout by running `git clean -fd`.

So I'm not really inclined to merge this one, to be honest, unless you make a good case for convincing me that the current approach is misguided. :)

noonedeadpunk:
> > How in the world I ended up with 3 commits here, when made only 1... I can't be more annoyed with the github flow then I am right now...
> 
> Doesn't look like this has much to do with GitHub; your change just wasn't based on `main`. Fixed now.

It was based of on my main, which I likely didn't sync somehow, despite 2 other PR were synced... So I don't know really, and there's no way to tell now either.

> So I'm not really inclined to merge this one, to be honest, unless you make a good case for convincing me that the current approach is misguided

Well, if required approach is to use CLI and VIM - then it might be fine. But IDEs go really crazy with that:
Screenshot_20230629_155327
And in these 500+ changed files there's 1 really changed intentionally, that should be commited rather then cleaned up, but it's quite tough to see it visually.
Moreover, things like VScode become quite hungry on resources when it starts tracking 500 files and changes in them (and highlighting these changes).

Also I don't fully understand why `git clean -fd` would be used. From what I see `tox` takes care of cleaning up old content in a quite reliable way. For me it makes more harm then help, and frankly speaking I hardly ever saw such behaviour in any project. It's kinda default, that things which should never be accidentally commited or not intended to be pushed ever, are added to .gitignore.
Just checked couple of projects outside of openstack, like [ansible](https://github.com/ansible/ansible/blob/devel/.gitignore), [ceph](https://github.com/ceph/ceph/blob/main/.gitignore), [mattermost](https://github.com/mattermost/mattermost/blob/master/.gitignore), [rabbitmq](https://github.com/rabbitmq/rabbitmq-server/blob/main/.gitignore) - all of them ensure that such build stuff won't raise and won't be accidentally commited. So that's smth that a developer expects when interacting with a project.

noonedeadpunk:
Sure, works for me, thanks!

PR #201, opened 2023-07-04
Bring back external link checking

fghaas:
Our automated checks for dead links haven't been functional, or at least not reliable as far as external links are concerned, for a while now. Since we can't really make them work reliably with the htmlproofer plugin because of https://github.com/manuzhang/mkdocs-htmlproofer-plugin/issues/63, try an alternate approach with linkchecker. 

- build: Drop envdir from testenv
- build: Replace htmlproofer plugin with linkchecker


colder-is-better (approved):
LGTM!

PR #202, opened 2023-07-04
build: Add workflow_dispatch event trigger to check.yml

fghaas:
Adding this event trigger is necessary to allow us to invoke the
workflow run manually.

References:
https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch


colder-is-better (approved):
The empty set always troubles mathematicians, so now that it's gone everything is looking good to me!

PR #203, opened 2023-07-04
build: Allow check job to continue when "tox -e check" fails

fghaas:
Without this change, the execution of the check job stops when `tox -e check` returns nonzero, meaning the report file is never uploaded.

Needless to say, we are very much interested in this artifact *particularly* when the link check fails, so use `continue‑on‑error: true` to make sure it is always uploaded.

Reference:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,6 +1,6 @@
 ---
 jobs:
-  build:
+  check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


colder-is-better (approved):
This is looking good to me!

PR #205, opened 2023-07-04
fix: Work around HTTP 500 at ceph.io (again)

fghaas:
This was mixed up in two prior commits, and then missed in review.

This reverts commit ecb633f27b06557cee9162935e4d4c25a8328863.

### DIFF
--- a/docs/background/disaster-recovery.md
+++ b/docs/background/disaster-recovery.md
@@ -22,7 +22,7 @@ background, and why you should consider enabling it.
 ## What it is
 
 The *Disaster Recovery* (DR) feature is available via the {{gui}} and
-applies to servers and volumes that use our [Ceph](https://ceph.io)
+applies to servers and volumes that use our [Ceph](https://ceph.io/en/)
 backend. That would be **all** servers but the ones of the `s`
 [flavor](../../reference/flavors/#compute-tiers).
 


docs/background/disaster-recovery.md
```diff
@@ -22,7 +22,7 @@ background, and why you should consider enabling it.
 ## What it is
 
 The *Disaster Recovery* (DR) feature is available via the {{gui}} and
-applies to servers and volumes that use our [Ceph](https://ceph.io)
+applies to servers and volumes that use our [Ceph](https://ceph.io/en/)
```
colder-is-better, line 25: When I visit https://ceph.io, I am automatically redirected to https://ceph.io/en anyway. Why not leave this kind of initiative to the remote server? I mean, https://ceph.io/en may stop working in the future. Why should we be the ones directly pointing to such a URL?

colder-is-better (commented):
I only have a question.

docs/background/disaster-recovery.md
fghaas: Simple: [the redirect is unreliable](https://github.com/citynetwork/docs/actions/runs/5456293589/jobs/9928912864#step:5:40). 

I already explained this in the commit message for ecb633f27b06557cee9162935e4d4c25a8328863, but then got the patch backwards.

colder-is-better (approved):

PR #206, opened 2023-07-05
build: Rewrite link checker, make it run on commit

fghaas:
Prior to this change, the dead links check with `linkchecker` could only be run post-merge. With this change, it checks for dead links on every push, as part of the normal CI testing via GitHub Actions.

* Make `linkcheck.sh` behave differently based on whether it's invoked as `linkcheck-local.sh` or `linkcheck-production.sh`.

* For `linkcheck-local.sh`, fire up a server on `localhost:8000`, and run `linkchecker` against that. Use that as part of the GitHub Actions `build` job, which runs on every push and PR.

* For `linkcheck-production.sh`, run `linkchecker` against <https://docs.cleura.cloud>. Invoke that from the `check` job, which runs on a schedule and can also be invoked manually.


colder-is-better (approved):
It is looking good!

PR #209, opened 2023-07-07
fix: Bump latest supported Kubernetes version for Gardener

fghaas:
The latest available Gardener-managed Kubernetes version in Cleura Cloud is no longer 1.24, but 1.26. Fix the relevant background and reference pages.


murwall (approved):
LGTM

colder-is-better (approved):

PR #213, opened 2023-07-13
feat: Mention community images in reference

noonedeadpunk:
We have recently switched our image rotation mechanism from
deactivating the image, which makes impossible any futher usage of the
image, to marking them as "comminuty" ones. This allows to use
already rotated image in case of need. Potential usecases are:
* Server rebuild
* New images somehow misbehave
* User will to use exact same image for another VM in the cluster

While usage of community images is inadvisable, we still should
mention their existance.

### DIFF
--- a/docs/reference/images/index.md
+++ b/docs/reference/images/index.md
@@ -50,3 +50,20 @@ All public images available in {{brand}} support the following image properties:
 
 [Other properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html) may also be set on individual images.
 In particular, {{brand}} aims to set image properties according to the [metadata standard](https://docs.scs.community/standards/scs-0102-v1-image-metadata/) defined by the [Sovereign Cloud Stack](https://scs.community/) (SCS) initiative.
+
+## Community images
+
+At {{brand}}, we regularly update and rotate our images to always provide
+secure public images.
+
+During rotation, we change an image's visibility from `public` to `community`,
+while keeping its image name. This enables tools like
+[Heat](https://docs.openstack.org/heat/) or [Terraform](https://www.terraform.io/)
+to pass validation checks without attempting to alter environments.
+
+You can retrieve an image's original build date (for images with both
+`public` and `community` visibility) by checking its `build_date` tag or
+`image_build_date` property.
+
+> Usage of community images is not recommended and is always upon
+> full responsibility of the user.


fghaas:
@noonedeadpunk, I believe this documentation update is no longer in line with how we manage images. Is that correct? If so, should we close this one?

fghaas:
Okay, can you please squash those two commits into one, and also rebase them on current `main`, before we review?

noonedeadpunk:
done

docs/reference/images/index.md
```diff
@@ -48,3 +48,19 @@ All public images available in {{brand}} support the following image properties:
 * `os_version=${VERSION_ID}`: operating system version, such as `os_version=22.04`
 
 [Other properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html) may also be set on individual images.
+
+### Community images
+
+At {{brand}} we regulary update and rotate our images to always provide
```
fghaas, line 54:
```suggestion
At {{brand}}, we regularly update and rotate our images to always provide
```

docs/reference/images/index.md
```diff
@@ -48,3 +48,19 @@ All public images available in {{brand}} support the following image properties:
...
+secure public images.
+
+During rotation old images are moved from `public` to `community` visability while
+keeping same images names. This enables tools like OpenStack Heat or HashiCorp
+Terraform to pass validation checks without attempting to alter environments.
```
fghaas, line 59:
```suggestion
During rotation, we change an image's visibility from `public` to `community`, while
keeping its image name. This enables tools like [Heat](https://docs.openstack.org/heat/)
or [Terraform](https://www.terraform.io/) to pass validation checks without attempting to alter environments.
```

docs/reference/images/index.md
```diff
@@ -48,3 +48,19 @@ All public images available in {{brand}} support the following image properties:
...
+
+You still can check image original build date for both `public` and `community` images
+by checking `build_date` tag or `image_build_date` property that will have date in
+`YYYY-mm-dd` format.
```
fghaas, line 63:
```suggestion
You can retrieve an image's original build date (for both `public` and `community` images)
by checking its `build_date` tag or `image_build_date` property.
```

fghaas (changes requested):
Thanks, I have a few suggestions for this one. :)

noonedeadpunk:
> Not quite; topic branch wasn't rebased on current state of `main`. :) I've taken care of that now.

Ah, true, I've fetched upstream, rebased main on top of upstream/main, pushed main to origin/main, but then didn't rebase that on top of origin/main... When trivial rebase requires 5 steps to execute, it's not hard for things to slip :(

fghaas:
I don't quite see 5 steps?

```bash
git checkout bugfix/community_images
git fetch upstream main:main
git rebase main
```

But if even 3 is too much, a clean rebase on current upstream `main` amounts to two clicks here in the web interface. :)

PR #214, opened 2023-07-19
fix: Remove information about object storage quotas

fghaas:
We no longer have any per-bucket object or data quotas defined in our
radosgw instances. Thus, remove the mentions of these quotas from the
reference section.


fghaas:
This supersedes #212. Courtesy tag for @pejotes.

colder-is-better (approved):
LGTM!

PR #216, opened 2023-08-14
chore: Link to Markdown paths internally

fghaas:
Newer versions of MkDocs rightfully flag links that point to relative
or absolute URL paths in the rendered tree as issues. Update those
links to pointers to their respective Markdown sources instead, so
that MkDocs can on its own figure out the correct links when rendering.


colder-is-better (approved):
Every replacement --and I mean, each and every single one of them-- looks good to me.

PR #217, opened 2023-08-14
fix: STO1HS IPv6 is being worked on

packettoobig:
Might have been a "feat:" PR, not sure.


fghaas (approved):
Fixed the commit message, to unbreak the build check. Looks good to me now.

PR #218, opened 2023-08-16
build: Post pull request comments on build check failure

fghaas:
Without this change, when a PR results in a build failure, the onus is on the PR submitter to find the failed build, and the corresponding build log. That is somewhat tedious and not particularly intuituive.

Instead, use a separate GitHub Action that posts a comment back to the PR thread, containing a direct link to the failed build log. This should hopefully make it easier for PR submitters to find out what exactly their submitted change breaks.

Reference:
https://github.com/quipper/comment-failure-action/pull/6

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -1,0 +1,20 @@
+# Post a comment back on a PR thread, in case an associated workflow
+# failed. Note that comment-failure-action is a no-op in case the
+# workflow was not invoked in response to a PR. In other words, this
+# action only creates comments for workflows configured with
+# "on: pull_request".
+---
+jobs:
+  comment-failure:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: quipper/comment-failure-action@v0.1.1
+
+name: comment
+
+'on':
+  workflow_run:
+    types:
+      - completed
+    workflows:
+      - build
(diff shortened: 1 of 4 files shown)


.github/workflows/comment.yml
```diff
@@ -0,0 +1,20 @@
+# Post a comment back on a PR thread, in case an associated workflow
+# failed. Note that comment-failure-action is a no-op in case the
+# workflow was not invoked in response to a PR. In other words, this
+# action only does creates a comments for workflows configured with
```
colder-is-better, line 4:
```suggestion
# action only createscomments for workflows configured with
```

colder-is-better (commented):
A small nit is all I have!

fghaas:
Thanks for catching that; I've fixed that nit. Also, please note that this is one of those things that we can only _actually_ test once they've landed in `main`, so apologies in advance in case this happens to break anything.

colder-is-better (approved):
LGTM!

colder-is-better:
> 



> Thanks for catching that; I've fixed that nit. Also, please note that this is one of those things that we can only _actually_ test once they've landed in `main`, so apologies in advance in case this happens to break anything.

Makes sense, yes.

fghaas:
Oddly enough, GitHub marked this one as Closed (as if it hadn't been merged). For the record, this was merged and then backed out again in #219.

PR #219, opened 2023-08-16
build: Drop the comment workflow

fghaas:
The workflow introduced in https://github.com/citynetwork/docs/pull/218, intended to facilitate pull requests by adding a comment in case the PR produced any build errors, failed on its first test. A build failure on https://github.com/citynetwork/docs/pull/217 did cause the action to fire, but it immediately failed with the following message:

```
Run quipper/comment-failure-action@v0.1.1 
No open pull requests found with packettoobig-patch-1 branch
```

Since a pull request from that branch obviously did exist, it follows that the action isn't working as intended, so drop it and rely on per-user build notifications, set via https://github.com/settings/notifications, instead.


fghaas:
@colder-is-better Apologies for the noise. 'Twas worth a shot. :)

colder-is-better (approved):

PR #227, opened 2023-08-21
build: Add validation parameters to mkdocs.yml

fghaas:
Since MkDocs 1.5, we can add a `validation` map to the configuration file, which we can use to upgrade validation `INFO` messages to the `WARNING` level — and this, in combination with the `--strict` setting in the `build` testenv, can flag such issues as errors. This is what we want, thus, set the values that MkDocs lists as "recommended settings for most sites (maximal strictness)".

Reference:
https://www.mkdocs.org/user-guide/configuration/#validation


colder-is-better (approved):
LGTM!

fghaas:
This is now merged and deployed, however any build checks on further PRs will henceforth fail, unrelated to *this* PR, because of #228. If that issue doesn't get fixed upstream shortly, we'll have to implement a workaround.

PR #229, opened 2023-09-01
fix: Add non-breaking spaces to brand names

fghaas:
We generally want "Cleura Cloud" to not break across a newline, so update the extra variable to include a non-breaking space.  Also, ensure that "Cleura Cloud Management Panel" and "Cleura Cloud REST API" only break into two halves, as opposed to just randomly.


colder-is-better (approved):
Magical spaces, nice!

PR #231, opened 2023-09-04
fix: Mention limit of 25 concurrently attached volumes

fghaas:
When launching a Nova server in Cleura Cloud, using one of the images marked as public in Glance, the guest operating system gets VirtIO block (`virtio-blk`) devices to work with. In the guest operating system, these images are named `/dev/vda` for the boot block device, and `/dev/vdb`, `/dev/vdc` etc. for attached volumes. Since VirtIO only enumerates devices up to `/dev/vdz`, this means it can only handle 26 devices in total, and thus a single server can only ever attach 25 volumes concurrently.

Until we have the option of running from public images configured with `virtio-scsi` (which does not have this limitation), state that 25 is the maximum number of volumes that can be concurrently attached to a single server.

### DIFF
--- a/docs/reference/limitations/openstack.md
+++ b/docs/reference/limitations/openstack.md
@@ -31,6 +31,11 @@ This means that you can run nested virtualization on servers booted from a CentO
 
 Furthermore, you must ensure that the Nova server [passes the `pcid` CPU feature flag](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#important-cpu-features-for-intel-x86-hosts) to nested guests.
 
+### Maximum attached volumes per server
+
+A Nova [server](../../howto/openstack/nova/new-server.md) in {{brand}} can concurrently attach a maximum of 25 persistent volumes.
+This is a limitation of the `virtio-blk` storage driver that ships as part of the guest operating system's kernel.
+
 ## Neutron
 
 ### Dynamic routing
--- a/tox.ini
+++ b/tox.ini
@@ -94,7 +94,7 @@ commands = yamllint {toxinidir}
 [testenv:markdownlint]
 envdir = {toxworkdir}/markdownlint
 deps =
-  pymarkdownlnt
+  pymarkdownlnt!=0.9.13
 commands = pymarkdown -c .pymarkdown.json scan -r docs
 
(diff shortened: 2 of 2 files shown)


fghaas:
@noonedeadpunk Do you have any objections to this one? I think it does reflect the current state of things correctly.

fghaas:
> Looks good (and Spartan) to me: Precise, concise, and exact -- like this comment.

"Laconic", indeed. :)

docs/reference/limitations/openstack.md
```diff
@@ -31,6 +31,11 @@ This means that you can run nested virtualization on servers booted from a CentO
 
 Furthermore, you must ensure that the Nova server [passes the `pcid` CPU feature flag](https://qemu-project.gitlab.io/qemu/system/qemu-cpu-models.html#important-cpu-features-for-intel-x86-hosts) to nested guests.
 
+### Maximum attached volumes per server
+
+A Nova [server](../../howto/openstack/nova/new-server.md) in {{brand}} can concurrently attach a maximum of 25 persistent volumes.
+This is a limitation of the `virtio-blk` storage driver that ships as part of the guest operating system's kernel.
+
```
noonedeadpunk, line 38: To be honest, I would mention a way to change disk bus to virtio-scsi with private image. While it's not a straigthforward way of doing things, usecase of having more then 25 volumes is quite specific as well. So might be good to say in 1 sentence that it's not a very hard limitation.

I'm also planning to rollout Fedora CoreOS images for magnum with virtio-scsi enabled, so that won't be a problem for k8s clusters spawned with magnum. And nobody care about disk performance in k8s world anyway I assume...

fghaas: > So might be good to say in 1 sentence that it's not a very hard limitation.

That's the point; you can't say this in 1 sentence so that it is meaningful to someone who knows what Glance image properties and virtio-scsi are, and how to create and upload images, while not being horribly confusing to someone who does not. So, no. We can write a full howto for this and then cross-reference that, but unless such a thing exists, I prefer to leave this as is.

Also, once *public* images with `virtio-scsi` are out, we can add a mention of that as an additional workaround.

PR #232, opened 2023-09-05
fix: Update version reference for Antelope

fghaas:
* Mention that Public Cloud is transitioning to Antelope (from Xena).
* Explain why we skip both Yoga and Zed, as opposed to just Yoga.
* Update the OpenStack version for Tky1 to Antelope.

### DIFF
--- a/docs/reference/versions/index.md
+++ b/docs/reference/versions/index.md
@@ -1,16 +1,12 @@
 # Service version matrix
 
-Services in {{brand}} are updated on a regular basis and
-on a rolling schedule.
+Services in {{brand}} are updated on a regular basis and on a rolling schedule.
 
-This section lists the cloud API service versions available in each
-{{brand}} region.
+This section lists the cloud API service versions available in each {{brand}} region.
 
-* [Public Cloud](public.md): versions running in our {{company}}
-  Public Cloud regions.
+* [Public Cloud](public.md): versions running in our {{company}} Public Cloud regions.
 
-* [Compliant Cloud](compliant.md): versions running in our
-  {{company}} Compliant Cloud regions.
+* [Compliant Cloud](compliant.md): versions running in our {{company}} Compliant Cloud regions.
 
 
 ## OpenStack Services
@@ -25,9 +21,6 @@ We return to the prior deployment schedule after the Antelope upgrade, and expec
 
 ## Ceph Services
 
-[Ceph major
-releases](https://docs.ceph.com/en/latest/releases/index.html#release-timeline)
-are also named, in alphabetical order, and occur on a roughly annual schedule.
+[Ceph major releases](https://docs.ceph.com/en/latest/releases/index.html#release-timeline) are also named in alphabetical order, and occur on a roughly annual schedule.
 
-{{brand}} currently runs Ceph
-[Quincy](https://docs.ceph.com/en/latest/releases/quincy).
+{{brand}} currently runs Ceph [Quincy](https://docs.ceph.com/en/latest/releases/quincy).
--- a/docs/reference/versions/index.md
+++ b/docs/reference/versions/index.md
@@ -15,14 +15,12 @@ This section lists the cloud API service versions available in each
 
 ## OpenStack Services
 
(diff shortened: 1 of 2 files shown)


fghaas:
This is marked as a Draft PR until the update actually completes, but preemptive reviews are OK. :)

docs/reference/versions/index.md
```diff
@@ -1,35 +1,26 @@
 # Service version matrix
 
-Services in {{brand}} are updated on a regular basis and
-on a rolling schedule.
+Services in {{brand}} are updated on a regular basis and on a rolling schedule.
 
-This section lists the cloud API service versions available in each
-{{brand}} region.
+vThis section lists the cloud API service versions available in each {{brand}} region.
```
colder-is-better, line 5: There's an extra "v" at the beginning of the sentence:
```suggestion
This section lists the cloud API service versions available in each {{brand}} region.
```

colder-is-better (commented):
All is looking good (there's only one stray character there...)

docs/reference/versions/index.md
fghaas: Oh, weird. Fixed now; no idea how that got in. Well spotted, thanks!

docs/reference/versions/index.md
```diff
@@ -1,35 +1,26 @@
...
 
 ## OpenStack Services
 
-[OpenStack releases](https://releases.openstack.org) are named, in
-alphabetical order, and occur on a six-month release schedule. In
-{{company}} Public Cloud we upgrade OpenStack releases annually; this
-means that we deploy every other OpenStack release and skip the
-intervening one.
+[OpenStack releases](https://releases.openstack.org) are named in alphabetical order, and occur on a six-month release schedule.
+In {{company}} Public Cloud we upgrade OpenStack releases annually; this means that we normally deploy every other OpenStack release and skip the intervening one.
```
noonedeadpunk, line 15: We can tell even that we're tracking SLURP releases, to make an expectation what next release will be.

docs/reference/versions/index.md
```diff
@@ -1,35 +1,26 @@
...
 
-{{brand}} currently runs OpenStack
-[Xena](https://releases.openstack.org/xena).
+{{brand}} currently runs OpenStack [Xena](https://releases.openstack.org/xena) and [Antelope](https://releases.openstack.org/antelope).
+The fact that {{brand}} has skipped the [Zed](https://releases.openstack.org/zed) release is due to a one-time upstream release policy change.
```
noonedeadpunk, line 18: We also skipped Yoga then. The reason why we skipped both is because we decided to align with SLURP schedule.

fghaas: That's what this says: in addition to the skipping the Yoga release (which is what we would would have done anyway, as the preceding paragraph explains), we also skipped Zed.

docs/reference/versions/index.md
noonedeadpunk: I'd say it does help. First, we can add a link to the governance decision explaining how SLURP is different. This resolution also describes periodity of SLURP releases. Then, we're mentioning, that we're doing only officially supported upgrades by the community and aiming to follow best practices rather then weirdly jumping through releases.

Also, I don't see much sense tracking down upstream docs, when we all know that next SLURP release will be 2024.1(Caraval) and it will marked as one as soon as it will become in Development (or Maintained - not sure when they add SLURP tag to the page tbh).

docs/reference/versions/index.md
colder-is-better: @fghaas I see it as an improvement. As a reader, I appreciate that level of detail. If anything, it shows that the people behind the scenes have good reasons for whatever they do or don't do.

fghaas: @colder-is-better Fair. I've squashed that fixup, and updated the PR branch. Please re-approve so we can merge this.

colder-is-better (approved):

fghaas:
Thanks @colder-is-better; I'll merge this when Mozilla fixes their HTTPS cert and the link check no longer breaks. 🥴

PR #233, opened 2023-09-20
feat: Add redirects

fghaas:
We should have the ability to change the names of Markdown sources and, hence, the URLs of rendered pages. However, in doing so we must make sure that existing links to our documentation do not break.

The [MkDocs `redirects` plugin](https://github.com/mkdocs/mkdocs-redirects) enables exactly that. This PR adds it to the dependencies in `tox.ini`, enables it in `mkdocs.yml`, and also adds a note to the contribution guide explaining how to add a redirect.

Finally, it makes use of that plugin by renaming a howto guide in the Account & Billing section, so that its URL matches its page title and navigation entry.

This change is meant to land *after* PR #234.


colder-is-better (dismissed):
LGTM also!

fghaas:
Oh this is weird. Just because `main` is now already up to 0a409faf93c945908512197ea2a1b6980a1a2fc3 as a result of #234 having merged, and "the merge-base changed after approval", you'll now need to re-approve. That's mighty silly, but can I ask you to approve one more time?

colder-is-better (approved):

fghaas:
Now it says the same thing _again?_ OK, that's a bug. (I didn't dismiss a thing; GitHub made that up by itself.)

Guess I'll have to actually update something, force-push to my topic branch, and then you re-approve _again._ Sigh...

fghaas:
Went with a different approach: enabled bypassing the branch restrictions, fast-forward merged offline, then pushed and re-disabled bypassing the branch restrictions.

colder-is-better:
I mean, when you **can** bypass restrictions, then why not? :)

fghaas:
Well the restrictions are there for a reason, and rules are meant to be for everyone. I maintain, though, that we did follow the spirit of the rule.

colder-is-better:
Uhm, totally agree (I was just joking :))

PR #234, opened 2023-09-20
build: Fix link checks

fghaas:
With the release of linkchecker 10.3.0, `tox -e check` has started to fail on *external* links in the documentation that happen to point to a target that is being redirected. The fix included here makes linkchecker ignore, specifically, its new `http-redirected` warning.

It also does fix *some* of the links that linkchecker flags when configured *not* to ignore `http-redirected`, but it does not make sense to change all such links — for about half of them, the redirect exists for a good reason.

### DIFF
--- a/.linkcheckerrc
+++ b/.linkcheckerrc
@@ -1,0 +1,10 @@
+[checking]
+recursionlevel=2
+
+[filtering]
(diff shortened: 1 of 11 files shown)


colder-is-better (approved):
Looking good!

PR #235, opened 2023-09-20
feat: Add Support category incl. Raising Issues article

AboveHero:
With this PR we would introduce the addition of a Support category in the "How-to Guides" section.
An article has been written to detail the different ways that entities of the different available support package plans can issue tickets to the Cleura Service Center.

Included are also three images (screenshots) of the three different categories of issue types, as well as an original image depicting our prioritization matrix.

The .pages file have been modified to add the Support category to the main navigation.

This is my first PR and attempt to modify/create content for the Cleura Docs, I hope I have followed the guidelines properly.

I noticed that in my commit I by force of habit capitalized the F in feat, I hope this is not an issue.


docs/howto/support/raising-issues.md
```diff
@@ -0,0 +1,111 @@
...
+- Invoice issues
+- Account issues
+- Cloud Management Portal issues
+
+#### Incidents
+
+When you issue a incident ticket with us you are reporting an **ongoing** incident.
+An incident is something that is hindering your normal operation, when an issue with our services is impacting your systems, customers or users.
+
+### Priority Matrix
```
fghaas, line 33: I think the information about the priority matrix should move to the bottom of the page. This subsection is not immediately relevant to _filing_ an issue, rather it explains how the issue will be prioritised once it _has_ been filed.

docs/howto/support/raising-issues.md
```diff
@@ -0,0 +1,111 @@
...
+---
+
+## Support Portal Ticket
+
+In the Cleura Service Center Portal there are three categories of request types, each with several issue types.
+
+The ticket menu works in a dynamic way which means that depending on your selections made throughout the menu you will have different optional and required fields where you can input information.
+Always try to include as much relevant information as possible, this will help us help you more efficiently.
+
+The portal is built to be user-friendly and should be fairly straight forward.
```
fghaas, line 74: That's a very subjective sentence. I'd suggest to just remove it.

```suggestion
```

fghaas (commented):
This is a good start. I have some thoughts; please see my inline remarks. :)

fghaas:
One more note: before you update this PR with the next force-push to your topic branch, please run `tox -e markdownlint` locally, and fix the issues it highlights. Thanks!

AboveHero:
Very good feedback. I see you are using variables quite a bit, I should've checked through the available ones when I wrote this article. Will address these comments and will make to sure to use variables where applicable and reasonable.

docs/howto/support/raising-issues.md
AboveHero: This assumption is not entirely correct. We do require the priority to be set by the customer on the issuance of Incident tickets (since the recent update to the portal), so understanding which priority to use is immediately relevant to the filing of an (incident) issue.

It used to be we required the selection of Urgency and Impact and then we set the priority based on the matrix, but we are now flipping that around and having the customer map this.

fghaas: Hmmm! Interesting approach. :) But yes, in that case we'll keep the matrix where it is.

AboveHero:
I absolutely agree that it is quite messy to duplicate information in terms of written form and visual representation and I am glad you have provided an alternative solution to this.

Your second point is also a good one, this should indeed be clarified.

docs/howto/support/raise-issues.md
```diff
@@ -0,0 +1,124 @@
...
+---
+
+## Support Portal Ticket
+
+In the {{support}} portal there are three categories of request types, each with several issue types.
+
+The ticket menu works in a dynamic way which means that depending on your selections made throughout the menu you will have different optional and required fields where you can input information.
+Always try to include as much relevant information as possible, this will help us help you more efficiently.
+
+1. Go to [{{support}} portal](https://{{support_domain}}).
```
fghaas, line 71: Do we want this link to point to the root of the domain, or to the `/servicedesk` path? (It's currently the latter everywhere else in the documentation.)

fghaas (commented):
I've taken the liberty to rebase this on current `main` one more time, and tidy up the commit message. I just have one more question — just an answer on that one will suffice; the fix is trivial either way so I can take of that myself. See my inline comment please.

docs/howto/support/raise-issues.md
AboveHero: Ah! It should be the servicedesk link, I must have missed that detail. Good eyes!

fghaas: No worries, fixed. :)

fghaas (approved):

PR #236, opened 2023-09-26
Explain Octavia's Prometheus endpoint

fghaas:
* Explain how one adds a metrics endpoint in Octavia (available from OpenStack Antelope forward)
* Add the availability of that feature to the feature support matrix.

### DIFF
--- a/docs/howto/openstack/octavia/metrics.md
+++ b/docs/howto/openstack/octavia/metrics.md
@@ -1,0 +1,151 @@
+---
+description: How to add a Prometheus metrics endpoint to a load balancer.
+---
+# Enabling load balancer metrics
+
+> This feature is available in select {{brand}} regions.
+> Please check the [feature support matrix](../../../reference/features/index.md) for details.
+
+Once you have created load balancers in {{brand}}, you can add a special *listener* that exposes load balancer metrics.
+This listener is meant to be used as a data source for an existing [Prometheus](https://prometheus.io/) monitoring system.
+
+## Prerequisites
+
+The {{gui}} does not support defining metrics endpoints, so you will have to work with the OpenStack CLI.
+[Enable it](../../getting-started/enable-openstack-cli.md) for the region you will be working in, and make sure you have the Python `octaviaclient` package installed.
+For that, use either the package manager of your operating system, or `pip`:
+
+=== "Debian/Ubuntu"
+    ```bash
+    apt install python3-octaviaclient
+    ```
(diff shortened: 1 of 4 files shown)


docs/howto/openstack/octavia/metrics.md
```diff
@@ -0,0 +1,124 @@
...
+    ```
+
+## Assumptions and scenario
+
+We assume you already have an [HTTPS-terminated load balancer](tls-lb.md) that forwards client requests to a back-end server pool.
+In our test scenario, the load balancer was accepting HTTPS requests for `whoogle.example.com` and forwarding them to a two-member pool, with servers each running a Docker container for [Whoogle Search](https://github.com/benbusby/whoogle-search).
+
+## Creating a metrics listener
+
+Here is `mylb`, the load balancer we used during testing...
```
colder-is-better, line 37:
```suggestion
Here is `mylb`, the load balancer we used during testing:
```
(I use the dots when presenting more than one console output box, in the context of a long sentence.)

colder-is-better (commented):
Everything looks good, kudos on keeping the Whoogle example, and I only have a sub-atomic nit.

docs/howto/openstack/octavia/metrics.md
fghaas: Point taken (see what I did there?)

fghaas:
I have now reversed the order of the two commits in this branch, and created a separate PR for the Antelope version update (#237).

docs/howto/openstack/octavia/metrics.md
```diff
@@ -0,0 +1,151 @@
...
++---------------------+--------+
+| Field               | Value  |
++---------------------+--------+
+| provisioning_status | ACTIVE |
++---------------------+--------+
+```
+
+## Testing the listener
+
+Once the listener reports as `ACTIVE`, you should be able to check its endpoint.
```
colder-is-better, line 117:
```suggestion
Once the listener reports as being `ACTIVE`, you should be able to check its endpoint.
```

(the above reads better --- at least for me)

colder-is-better (commented):
It looks good --- and I have nothing but a small nit.

docs/howto/openstack/octavia/metrics.md
fghaas: Fair point; fixed as suggested.

noonedeadpunk (approved):
LGTM

colder-is-better (approved):

PR #238, opened 2023-10-06
fix: Prometheus listeners now available in Kna1

fghaas:
With base images for the Octavia amphorae having been updated in Kna1, `--protocol PROMETHEUS` listeners are now available in that region.

This PR will remain a Draft until the images are actually rolled out, but is good for review.


colder-is-better (commented):
LGTM!

colder-is-better (approved):
Both catches, really nice!

PR #240, opened 2023-10-18
chore: Use material.extensions.emoji

fghaas:
Up until mkdocs-material 9.3, `materialx.emoji` was the correct way to refer to the emoji generator. As of 9.4, we must use `material.extensions.emoji`.

Fix `mkdocs.yml` accordingly.


colder-is-better (dismissed):
Looking good!

PR #241, opened 2023-10-20
Update the OpenStack Services matrix

colder-is-better:
The Antelope has landed on the Fra1 region, so we update the corresponding support matrix accordingly.

### DIFF
--- a/docs/reference/versions/public.md
+++ b/docs/reference/versions/public.md
@@ -2,17 +2,17 @@
 
 ## OpenStack Services
 
-|                                | Kna1     | Sto2   | Fra1  | Dx1   | Tky1     |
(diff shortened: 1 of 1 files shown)


colder-is-better:
> Looks good, but please fix the commit message and thus pass the CI checks.
> 
> (Hint: `tox -e gitlint`)

Boy, that's a chore!

fghaas:
Nah it's really a `fix` because the documentation is presently not correct, and this change makes it correct again. (See also #237)

colder-is-better:
> Nah it's really a `fix` because the documentation is presently not correct, and this change makes it correct again. (See also #237)

Strange. It felt like a chore (albeit, a tiny one) to me :/

PR #243, opened 2023-10-23
fix: Update the OpenStack Services matrix

colder-is-better:
The Antelope has landed in the Sto1HS region of the Compliant Cloud,
so we update the corresponding support matrix accordingly.


fghaas:
Rebased on current `main`.

PR #251, opened 2023-12-05
fix: Introduce "Cleura Container Orchestration Engine"

fghaas:
What used to be "Gardener in Cleura Cloud" is now, apparently, "Cleura Container Orchestration Engine". Introduce a separate configuration variable for that purpose.


fghaas:
We won't merge this. Closing.

PR #252, opened 2023-12-05
refactor: Introduce extra variable for container engine

fghaas:
Our container orchestration engine brand, "Gardener in Cleura Cloud", may conceivably change at some point in the future. If that does happen, being able to change a single variable would facilitate the brand update.

Thus, include a new such variable, `brand_container_orchestration`, and set it to "Gardener in Cleura Cloud".


colder-is-better (approved):
LGTM!

PR #256, opened 2023-12-12
build: Use git-committers plugin

fghaas:
Rather than using `git-authors` (whose rendering functionality is now [natively included in mkdocs-material](https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#document-authors), and which always wants to display a `mailto:` link containing authors' email addresses), use the `git-committers` plugin which instead links to contributors' GitHub accounts.

Since this relies on GitHub API calls that are rate-limited and potentially quite time-consuming during a local build, enable the
plugin only when running from GitHub Actions workflows.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 ---
 jobs:
   build:
+    env:
+      DOCS_ENABLE_COMMITTERS_PLUGIN: true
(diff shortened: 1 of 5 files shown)


fghaas:
We'll go with #257 instead (which we will update and merge once mkdocs-material 9.5.3 is out).

PR #257, opened 2023-12-12
build: Use mkdocs-material 9.5.3, drop git-authors theme override

fghaas:
What will probably become mkdocs-material 9.5.3 gives us the ability to again render authors' names without them turning into "mailto:" links. Thus, require that as the minimum version. Since it is now mkdocs-material that natively renders author information, we no longer need the theme override for main.html.

References:
https://squidfunk.github.io/mkdocs-material/setup/adding-a-git-repository/#document-authors
https://github.com/squidfunk/mkdocs-material/pull/6494


tox.ini
```diff
@@ -11,7 +11,7 @@ deps =
   mkdocs-git-revision-date-localized-plugin
   mkdocs-glightbox
   mkdocs-macros-plugin
-  mkdocs-material~=9.4.0
+  git+https://github.com/squidfunk/mkdocs-material@master
```
fghaas, line 14: @colder-is-better: Just a heads-up that this exists. I will update this line to `mkdocs-material>=9.5.3` once that version is released, and then I will mark this PR ready for review.

colder-is-better: Ack, thanks.

colder-is-better (approved):

PR #260, opened 2023-12-14
fix: Add missing descriptions to Nova how-to guides

fghaas:
Several of the how-to guides for OpenStack Nova were missing front
matter with a description field. Add that as appropriate.


colder-is-better (approved):
LGTM!

PR #261, opened 2023-12-14
fix: Update vGPU How-To guide in light of profiles

colder-is-better:
While creating a new vGPU server, besides using the OpenStack CLI, we can now work from the Cleura Cloud Management Dashboard. Considering that, we update the prose of this How-To guide and add some related screenshots.


docs/howto/openstack/nova/new-vgpu-server.md
```diff
@@ -5,70 +5,98 @@ In select
 and
 [{{brand_compliant}}](../../../reference/features/compliant.md#virtualization)
 regions, you can instantiate virtual machines --- or simply _servers_
---- with virtual GPU (vGPU) support. For that, you use the `openstack`
-CLI utility.
+--- with virtual GPU (vGPU) support. For that, you may either use the
+{{brand}} Dashboard or the `openstack` CLI utility.
```
fghaas, line 9: Please don't invent another term for CCMP.

```suggestion
{{gui}} or the `openstack` CLI utility.
```

docs/howto/openstack/nova/new-vgpu-server.md
```diff
@@ -5,70 +5,98 @@ In select
...
-    --key-name mykey \
-    --wait \
-    my-vgpu-server
-```
-
-Before you check whether the server does have virtual GPU support, we
-recommend assigning a floating IP to it.
+=== "{{gui}}"
+    Following the instructions in the [server creation How-To
+    guide](new-server.md), go ahead and instantiate a new cloud server.
```
fghaas, line 20: Parallelism: use the same terminology as in the CLI section.
```suggestion
    Following the instructions in the [guide on creating servers](new-server.md), go ahead and instantiate a new cloud server.
```

docs/howto/openstack/nova/new-vgpu-server.md
```diff
@@ -5,70 +5,98 @@ In select
...
+
+    ![Create new object](assets/new-vgpu-server/shot-02.png)
+
+    Regarding the image your new vGPU server will be based on, your only
+    option is `Ubuntu 22.04 NVGRID x86_64`.
+
+    ![Create new object](assets/new-vgpu-server/shot-03.png)
+
+    Notice that the disaster recovery mechanism is _not_ available for this
+    type of server.
```
fghaas, line 38: I think it would be a good idea to drive home what this means, like so:

```suggestion
    Notice that the disaster recovery mechanism is _not_ available for this
    type of server.
    You will not be able to [restore the server to a snapshot](restore-srv-to-snap.md).
```

fghaas (commented):
Looks good; just a few notes:

First, now that the CLI is no longer a hard requirement for spinning up a vGPU-enabled server, I would suggest to remove the sentence "Alternatively, connect to it using the `openstack` utility:" and the following console snippet. (GitHub doesn't let me suggest this in an inline review comment since that particular part isn't being touched by the patch.)

Also, please rebase on your patch on current `main`.

For everything else, see the inline comments.

docs/howto/openstack/nova/new-vgpu-server.md
colder-is-better: No, no new term.

docs/howto/openstack/nova/new-vgpu-server.md
colder-is-better: Yes.

docs/howto/openstack/nova/new-vgpu-server.md
colder-is-better: It is a good idea, yes.

fghaas (approved):
This looks great now; I'll go ahead and merge this. Thanks!

PR #264, opened 2023-12-18
fix: Drop :8080 port suffix from S3 how-to guides

fghaas:
Now that all our object storage services are listening on the regular HTTPS port (443), we no longer need to list the `:8080` suffix in our how-to guides for S3.
    
We still do need to keep the suffix in the how-to guides for Swift, since our `object-storage` endpoints in our service catalogs still do contain the `:8080` suffix.

### DIFF
--- a/docs/howto/object-storage/s3/credentials.md
+++ b/docs/howto/object-storage/s3/credentials.md
@@ -66,16 +66,16 @@ How exactly you do that depends on your preferred client:
       aws_secret_access_key <secret-key>
     aws configure set \
       --profile <region> \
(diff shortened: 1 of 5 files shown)


colder-is-better (commented):
LGTM!

pejotes:
This looks good to me as well.

fghaas:
Split the one commit into two since we're having to handle this change separately for the S3 and Swift APIs.

fghaas:
@colder-is-better This will need a re-review please, due to having been rebased on current `main`. Thanks!

colder-is-better (approved):

PR #265, opened 2023-12-19
fix: Drop :8080 port suffix from Swift how-to guides

fghaas:
After the service catalog endpoints for object-storage have been updated to drop the `:8080` port suffix, we can eliminate that suffix from the how-to guides for Swift.


fghaas:
Doesn't make a ton of sense to keep this one around indefinitely.

PR #266, opened 2024-01-02
fix: Add missing screenshot to Gardener upgrade How-To

colder-is-better:
A screenshot from the Gardener upgrade How-To guide is missing, so we make sure it's there. While at it, we refresh all screenshots and introduce slight modifications to prose due to minor UI changes.


fghaas (commented):
Looks great. Just one request: while I do concede that the emoji overlays over the screenshots are charming, I think they're a bit out of place. :) So let's please drop those.

colder-is-better:
Gone are the emojis!

fghaas (approved):
Thanks!

PR #273, opened 2024-01-19
fix: Add region decommission notes

colder-is-better:
There are two planned decommission events:
* 2024-04-30: region Tky1
* 2024-06-30: region Dx1

### DIFF
--- a/docs/reference/features/public.md
+++ b/docs/reference/features/public.md
@@ -7,7 +7,13 @@
 > :material-logout: Feature is deprecated, being phased out
 >
 > :material-close: Feature is not available
-
+> ## Sunset dates
+>
+> Please note that sunset dates have been set for the following {{brand}} regions:
+>
+> * The **Tky1** region will no longer be available after 2024-04-30.
+>
+> * The **Dx1** region will no longer be available after 2024-06-30.
 
 ## Virtualization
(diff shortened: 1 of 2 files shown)


docs/reference/features/public.md
```diff
@@ -7,7 +7,12 @@
 > :material-logout: Feature is deprecated, being phased out
 >
 > :material-close: Feature is not available
-
+>
```
fghaas, line 10: This `>` character before an empty line doesn't look right.

docs/reference/features/public.md
```diff
@@ -7,7 +7,12 @@
...
+> ## Planned decommissions
+>
+> Region Dx1 will be decommissioned on 2024-06-30.
+>
+> Region Tky1 will be decommissioned on 2024-04-30.
```
fghaas, line 15: Could we perhaps be a little more polite to readers? Something like this?

```suggestion
> ## Sunset dates
>
> Please note that sunset dates have been set for the following {{brand}} regions:
>
> * The **Tky1** region will no longer be available after 2024-04-30.
> * The **Dx1** region will no longer be available after 2024-06-30.
```

fghaas (commented):
I'm a bit unhappy with the suggested language, and have suggested alternate phrasing.

fghaas (approved):

PR #275, opened 2024-02-08
fix: Update shoot management guide due to new kubeconfig options

colder-is-better:
Users of existing shoot clusters with Kubernetes 1.26 can choose between static and certificate-based kubeconfig files. On the other hand, users of new shoot clusters with Kubernetes 1.26 or any shoot cluster with Kubernetes 1.27 and higher, can only use certificate-based kubeconfig files. We update the How-To guide, refreshing screenshots and modifying the prose accordingly.


docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -17,16 +17,27 @@ With the advent of Kubernetes 1.27, you can access your shoot cluster using a *c
...
 * A **static** kubeconfig contains a [static token](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file), provides `cluster-admin` privileges to the corresponding shoot cluster, and never expires.
-In {{brand}}, up until and including Kubernetes 1.26, you could access your shoot cluster by generating a static kubeconfig.
+In {{brand}}, up until and including Kubernetes 1.26, you can access your shoot cluster by generating a static kubeconfig.
 
 * A **certificate-based** kubeconfig is valid for a predefined amount of time, during which it can be used for accessing the corresponding shoot cluster with `cluster-admin` privileges.
 The credentials associated with this type of kubeconfig are client certificates with time-limited validity.
-Beginning with Kubernetes version 1.27, for new clusters you are able to issue certificate-based kubeconfig files only.
+Beginning with Kubernetes version 1.27, for new Gardener-based clusters with Kubernetes 1.27 or later, you can issue certificate-based kubeconfig files only.
+
+> For clusters with Kubernetes version 1.26, you can issue a static *or* certificate-based kubeconfig.
```
fghaas, line 26: I think we need to clarify here that this is only possible for *pre-existing* clusters; a *newly created* 1.26 cluster will always use a cert-based kubeconfig. No?

fghaas (commented):
Looks great. Just one small question (please see the inline comment).

docs/howto/kubernetes/gardener/kubectl.md
colder-is-better: Ah, yes, that is my understanding also. I'm adding this clarification.

fghaas (approved):
This looks good to me now, thanks!

PR #278, opened 2024-03-06
fix: Update HowTo in light of worker subnet changes

colder-is-better:
While creating a new Gardener Shoot cluster, we now have more flexibility regarding the worker nodes subnet. Thus, we update the HowTo prose and add new screenshots.


docs/howto/kubernetes/gardener/create-shoot-cluster.md
```diff
@@ -25,11 +25,28 @@ At the top right-hand side of the central pane, click on *Create Kubernetes cl
 A new pane named *Create {{k8s_management_service}} Shoot Cluster* slides over from the right-hand side of the browser.
 Type in a name for the new shoot cluster, and select a region.
 Also, select the version of Kubernetes the new cluster will be running.
-In the example below, we have chosen version 1.28.6, which, at the time of writing, was the latest supported in {{brand}}.
+In the example below, we have chosen version 1.28.7, which, at the time of writing, was the latest supported in {{brand}}.
 
 !["Create {{k8s_management_service}} Shoot Cluster" panel, showing options to type in a cluster name, select region and Kubernetes version](assets/create-shoot-1.png)
 
-Scroll down a bit until you bring the *Worker Groups* section into full view.
+Next, you may accept the proposed CIDR for the worker nodes or type in the one you prefer.
```
fghaas, line 32: I believe we need a comma before "or" here, and also I don't think I like the phrase "the CIDR". CIDR is Classless Inter-Domain Routing; the proper term here would be "the CIDR network address" or some such.

Now I understand that the incorrect term is used in CCMP, so this adds potential confusion, but maybe we can get away with something like this:
```suggestion
Next, you may accept the proposed network address (in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation) for the worker nodes, or type in the one you prefer.
```
What do you think?

docs/howto/kubernetes/gardener/create-shoot-cluster.md
```diff
@@ -25,11 +25,28 @@ At the top right-hand side of the central pane, click on *Create Kubernetes cl
...
+
+![Accept the proposed CIDR for the worker nodes or type in the CIDR you prefer](assets/create-shoot-1.1.png)
```
fghaas, line 34: Here in the alt text, I'd say we can definitely use "proposed CIDR network address".

docs/howto/kubernetes/gardener/create-shoot-cluster.md
```diff
@@ -25,11 +25,28 @@ At the top right-hand side of the central pane, click on *Create Kubernetes cl
...
+
+Alternatively, click on the drop-down menu named *Network for worker nodes* and select a pre-existing menu, one with a subnet and a router.
```
fghaas, line 36: I believe you meant to include the word "item" after "menu" here (or maybe say "item" _instead of_ "menu").

docs/howto/kubernetes/gardener/create-shoot-cluster.md
```diff
@@ -25,11 +25,28 @@ At the top right-hand side of the central pane, click on *Create Kubernetes cl
...
+
+![Select a pre-existing network](assets/create-shoot-1.2.png)
+
+Then, define a new subnet for the worker nodes by typing in a CIDR.
+If the new CIDR defines a subnet that overlaps with one of the subnets of the network you just selected, you will see an error message in red.
```
fghaas, line 41: Again, the "a CIDR" problem. And I believe we can condense the second sentence a bit:
```suggestion
Then, define a new subnet for the worker nodes by typing in a CIDR network address.
If the new subnet overlaps with one of the subnets of the network you just selected, you will see an error message in red.
```

docs/howto/kubernetes/gardener/create-shoot-cluster.md
colder-is-better: Even though I know what CIDR means, I admit I got carried away by the CCMP wording. So, yes, what you propose is 100% accurate, much betterer (sic), and we should go with that.

docs/howto/kubernetes/gardener/create-shoot-cluster.md
colder-is-better: Ugh, no. What I mean is that there's a drop-down menu named "Network for worker nodes" (that's the name of the drop-down menu, I kid you not). Also, I should replace the "pre-existing menu" with "pre-existing network". I'll modify the whole sentence.

docs/howto/kubernetes/gardener/create-shoot-cluster.md
colder-is-better: Yes.

docs/howto/kubernetes/gardener/create-shoot-cluster.md
```diff
@@ -25,11 +25,29 @@ At the top right-hand side of the central pane, click on *Create Kubernetes cl
...
 
 !["Create {{k8s_management_service}} Shoot Cluster" panel, showing options to type in a cluster name, select region and Kubernetes version](assets/create-shoot-1.png)
 
-Scroll down a bit until you bring the *Worker Groups* section into full view.
+Next, you may accept the proposed network address (in [CIDR notation](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing#CIDR_notation)) for the worker nodes, or type in the one you prefer.
+
+![Accept the proposed CIDR network address for the worker nodes or type in the CIDR network address you prefer](assets/create-shoot-1.1.png)
+
+Alternatively, click on the drop-down menu at the right-hand side of "Network for worker nodes".
+From the available menu items select a pre-existing network -- as long as it has a subnet and a router.
```
fghaas, line 37: This still sounds a bit clumsy to me. How about this:
```suggestion
From the available menu items, choose a pre-existing network.
Be sure to select one that has an assigned subnet, and is connected to a router.
```

colder-is-better: Betterer, yes!

fghaas (approved):
This looks good to me now. The GitHub Actions workflow is due to LinkChecker complaining that an upstream documentation source is currently unavailable, so that's not a "real" merge block.

fghaas:
Merged. Thank you!

PR #279, opened 2024-03-08
fix: Unpublish HowTo guide on vGPU-powered servers

colder-is-better:
Due to specific changes in vGPU availability, this guide should not be available for now. At the same time, all references to vGPUs should go.


fghaas:
Removing just the how-to isn't sufficient here; we also need to fix the vGPU mentions in the feature reference (and any other cross-references to vGPU features).

colder-is-better:
@fghaas Ah, yes, there's a lot more :/ Working on it.

fghaas (approved):

PR #281, opened 2024-03-21
fix: Update Shoot management guide in light of new Monitoring tab

colder-is-better:
When bringing Shoot details into view, we now have a new tab labeled "Monitoring". We update relevant screenshots with newer versions that have this new tab. The prose does not have to be updated.


fghaas (approved):
Looks good to me, thank you!

PR #289, opened 2024-05-17
fix: Correct and enhance the object locking guide

revij:
This pull request addresses and fixes some issues in our object locking guide.

e5fbef5526675e6b2fcf3820df77db8516d6e5b9 - Updated the object locking guide to correctly reference the "put-object-lock-configuration" command instead of "put-object-retention" for setting default object lock configuration on a per-bucket level.

f02b32466bbd04c07c1a30c2be0aa0c2a602c6ef - Corrected references of the "put-legal-hold" command, replacing it with the correct "put-object-legal-hold" command. Also included an alternative method to set legal hold for multiple objects using the AWS CLI. 

e5e70dc4c538d1d044d6ece6861e2af0ef70c7f8 - Added a missing section on configuring object lock per object, including commands to set/update the retention mode/period for a single object in a lock-enabled bucket.

f8bf4f3785cb41a1a204799e710e2d8379bbe71f - Updated the mc CLI command syntax for object locking references and included a notice on how to remove the legal hold from an object.

### DIFF
--- a/docs/howto/openstack/cinder/encrypted-volumes.md
+++ b/docs/howto/openstack/cinder/encrypted-volumes.md
@@ -20,9 +20,9 @@ $ openstack volume type list
 +--------------------------------------+-----------------------+-----------+
 | ID                                   | Name                  | Is Public |
 +--------------------------------------+-----------------------+-----------+
-| a479a6b0-b283-41a5-b38b-5b08e7f902ca | volumes_hdd_encrypted | True      |
+| a479a6b0-b283-41a5-b38b-5b08e7f902ca | cbs-encrypted         | True      |
 | d9dfa98a-238d-4ca0-9abf-701fceb05623 | __DEFAULT__           | True      |
-| 86796611-fb12-4628-b6b1-e09469e301d7 | volumes_hdd           | True      |
+| 86796611-fb12-4628-b6b1-e09469e301d7 | cbs                   | True      |
 +--------------------------------------+-----------------------+-----------+
 ```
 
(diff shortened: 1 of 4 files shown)


docs/howto/object-storage/s3/object-lock.md
```diff
@@ -204,7 +204,7 @@ use the following command:
 
     ```bash
     aws --profile <region> \
-      s3api put-legal-hold \
+      s3api put-object-legal-hold \
```
fghaas: Since the S3 API documentation says [legal holds can only be placed on individual object versions, not whole objects](https://docs.aws.amazon.com/AmazonS3/latest/userguide/batch-ops-legal-hold.html), do we want to add a `--version-id` option here?

fghaas (commented):
A few minor notes.

Also, you will want to go over your commit messages with `git rebase -i main` and apply the `reword` option to some of them. It looks like you made several changes to your topic branch between starting to work on your changes and pushing them, which means that some of the commit messages no longer reflect the change that their commit actually makes.

docs/howto/object-storage/s3/object-lock.md
revij: Indeed it is probably better to include `--version-id` since if one does not specify it, it will only apply it to the **last** version of that object. Fixed on commit [9866793](https://github.com/citynetwork/docs/pull/289/commits/9866793cceefa6b952740f844401d36c676aaec4).

fghaas:
Sorry, messed up a push and accidentally closed this one as a result. Follow-up PR, rebased on current `main`, is in PR #291.

PR #292, opened 2024-06-18
fix: Explain that Cinder QoS limits are only enforced in Public Cloud

fghaas:
The `cbs` and `cbs-encrypted` volume types only come with QoS limits in Cleura Public Cloud. Explain this fact more clearly.


colder-is-better (approved):
Looks good, thanks.

damiandabrowski5 (approved):

PR #294, opened 2024-07-02
fix: Refresh Magnum cluster creation HowTo

colder-is-better:
There are new public templates available for creating Magnum clusters, so we make sure the corresponding HowTo guide is up-to-date.


docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -71,82 +70,99 @@ package manager of your operating system or `pip`:
...
+    +--------------------------------------+---------------------------------------------+------+
     ```
 
     Select the template you want by setting the corresponding `uuid`
     value to the  `CLUSTER_TMPL` variable:
 
-    ```bash
-    CLUSTER_TMPL="f9e1a2ea-b1ff-43e7-8d1e-6dd5861b82cf" # just an example
+    ```console
+    $ CLUSTER_TMPL="9ca03308-996e-4eaa-b507-5730dcc19fcc" # just an example
```
fghaas, line 94: I think now that we have defined public template names that should be identical across all regions, we can use the template name instead of the UUID and also drop the "just an example" bit:
```suggestion
    $ CLUSTER_TMPL="Kubernetes v1.24.16 on Fedora-coreos 37"
```

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -71,82 +70,99 @@ package manager of your operating system or `pip`:
...
     ```
 
     Then, list all available keypairs...
 
-    ```bash
-    openstack keypair list
-    ```
+    ```console
+    $ openstack keypair list
```
fghaas, line 101: I believe we can drop the extraneous blank line here.

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -71,82 +70,99 @@ package manager of your operating system or `pip`:
...
     ```
 
     The cluster creation process takes some time to complete, and while
     you are waiting, you can check if everything is progressing smoothly:
 
-    ```bash
-    openstack coe cluster list -c status
+    ```console
+    $ openstack coe cluster list -c status
+
```
fghaas, line 144: Here, too.

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -71,82 +70,99 @@ package manager of your operating system or `pip`:
...
+    +--------------------+
     ```
 
     If everything is going well, the message you will get will be
     `CREATE_IN_PROGRESS`. When Magnum has finished creating the cluster, the
     message will be `CREATE_COMPLETE`.
 
+    ```console
+    $ openstack coe cluster list -c status
+
```
fghaas, line 158: And here.

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -167,63 +183,62 @@ get detailed information about it.
 === "OpenStack CLI"
     To list all available Kubernetes clusters, type:
 
-    ```bash
-    openstack coe cluster list
-    ```
+    ```console
+    $ openstack coe cluster list
```
fghaas, line 188: And here.

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -167,63 +183,62 @@ get detailed information about it.
...
     ```
 
     For many more details on a specific cluster, note its name and run a
     command like this:
 
-    ```bash
-    openstack coe cluster show bangor
-    ```
+    ```console
+    $ openstack coe cluster show $CLUSTER_NAME
```
fghaas, line 204: And here.

fghaas (commented):
Looks very good to me! I have a few nits, and a question/suggestion: now that the public templates finally come with the master load balancer enabled by default, should we perhaps suggest/recommend to configure Magnum clusters with 3 control plane (master) nodes?

(Consider that customers can change the number of worker nodes in a running cluster, but not the number of control plane nodes.)

fghaas:
Oh, and importantly: please rebase your topic branch on current `main`.

docs/howto/openstack/magnum/new-k8s-cluster.md
colder-is-better: Well, I know those are not really there, but I usually add them for improved readability. On the other hand, we can definitely **not** add those extra blank lines, and no one will complain.

colder-is-better:
I updated the prose for that load balancer in the new templates.

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -66,87 +78,129 @@ package manager of your operating system or `pip`:
     openstack coe cluster create \
         --cluster-template $CLUSTER_TMPL \
         --keypair $KEYPAIR \
+        --docker-volume-size PERSIST_VOL_SIZE \
```
fghaas, line 81: Missing `$`, I think.
```suggestion
        --docker-volume-size $PERSIST_VOL_SIZE \
```

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -66,87 +78,129 @@ package manager of your operating system or `pip`:
...
 
-    ```bash
-    KEYPAIR="husavik" # again, this is just an example
+    ```console
+    $ KEYPAIR="lefkanti" # this is just an example
     ```
 
+    Besides the OS disk, all cluster nodes have an extra disk for permanently storing application data.
+    The size of this extra disk is specified during cluster creation via the `--docker-volume-size` parameter.
+    Staying loyal to the theme of setting parameters via environment variables, decide on the size of this extra volume like so:
```
fghaas, line 130:
```suggestion
    Staying loyal to the theme of setting parameters via shell variables, decide on the size of this extra volume like so:
```

fghaas, line 130:
```suggestion
    Staying loyal to the theme of setting parameters via shell variables, decide on the size of this extra volume like so:
```

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -66,87 +78,129 @@ package manager of your operating system or `pip`:
...
+    worker nodes *after* the cluster is created. So, to start the new cluster
+    with 3 master nodes and 5 worker nodes, type the following:
+
+    ```console
+    $ openstack coe cluster create \
+        --cluster-template "$CLUSTER_TMPL" \
+        --keypair $KEYPAIR \
+        --docker-volume-size $PERSIST_VOL_SIZE \
+        --master-count 3 \
+        --node-count 5 \
```
fghaas, line 168: It's not *completely* logical to me why we want to set the persistent volume size with a variable, while just using int literals for the master and worker node count. Is there a specific reason to go with the variable, rather than just `50`?

docs/howto/openstack/magnum/new-k8s-cluster.md
```diff
@@ -167,63 +221,59 @@ get detailed information about it.
...
     | status_reason        | None                                                                      |
-    | created_at           | 2022-11-14T07:32:02+00:00                                                 |
-    | updated_at           | 2022-11-14T07:37:26+00:00                                                 |
-    | coe_version          | v1.18.6                                                                   |
-    | labels               | {'kube_tag': 'v1.18.6', 'heat_container_agent_tag': 'train-stable'}       |
+    | created_at           | 2024-07-05T12:00:41+00:00                                                 |
+    | updated_at           | 2024-07-05T12:08:52+00:00                                                 |
+    | coe_version          | v1.24.16-rancher1                                                         |
+    | labels               | {'container_runtime': 'containerd', 'cinder_csi_enabled': 'True',         |
+    |                      | 'cloud_provider_enabled': 'True', 'docker_volume_size': '20', 'kube_tag': |
```
fghaas, line 254: Note: when the useless `docker_volume_size` *label* disappears from the cluster template, we'll need to update this bit.

fghaas (commented):
Looks good; I have a few thoughts to add.

docs/howto/openstack/magnum/new-k8s-cluster.md
colder-is-better: Ouch, yes :/

docs/howto/openstack/magnum/new-k8s-cluster.md
colder-is-better: Ah, yes.

docs/howto/openstack/magnum/new-k8s-cluster.md
colder-is-better: Well, we start with shell variables to show what a _basic_ cluster creation command looks like. A bit later during our demonstration, we make a somewhat parenthetical comment on the `--master-count` and `--node-count` parameters, and we give a quick but specific example of creating a cluster with `3` masters and `5` workers. We _specifically_ mention `3` and `5`, so introducing shell variables for _that_ particular example wouldn't make much sense, I believe.

fghaas:
I think this looks good now ­— is there anything you'd like to add, still?

colder-is-better:
> I think this looks good now ­— is there anything you'd like to add, still?

For the time being, no, I cannot think of something. I will have to add a comment on specifying the docker volume size from the CCMP, though, as soon as this feature is ready.

fghaas (approved):

colder-is-better:
> This looks good to me now. @colder-is-better, do you plan to add any more changes, or do you consider this one ready to be merged?

I do not have anything else to add/change, so we can go ahead with the merging.

PR #295, opened 2024-07-29
fix: Remove all references to Dx1 region

colder-is-better:
We are now past the final deadline for customers to move out of Dx1, so we proceed with removing all relevant references from the public-facing documentation.

### DIFF
--- a/docs/reference/versions/public.md
+++ b/docs/reference/versions/public.md
@@ -1,28 +1,24 @@
 # Public Cloud
 
-> ## Sunset dates
->
-> Please note that the **Dx1** region will no longer be available after 2024-07-21.
-
 ## OpenStack Services
 
-|                                | Kna1     | Sto2     | Fra1      | Dx1       |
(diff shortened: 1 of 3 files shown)


colder-is-better:
For me, the (local) Markdown linter tests succeed:

```
$ tox -e markdownlint
markdownlint: commands[0]> pymarkdown -c .pymarkdown.json scan -r docs
  markdownlint: OK (1.21=setup[0.02]+cmd[1.19] seconds)
  congratulations :) (1.26 seconds)
```

And yet, the checks fail with `markdownlint: FAIL code 1`.

docs/reference/versions/public.md
```diff
@@ -1,28 +1,22 @@
-# Public Cloud
```
fghaas, line 1: Don't remove this heading. (This is what `pymarkdown` is complaining about.)

colder-is-better: Got it, thanks!

fghaas (approved):

PR #298, opened 2024-10-02
fix: Update Gardener kubectl How-To guide

colder-is-better:
From Kubernetes 1.27 on, we support certificate-based kubeconfig files only. For that, we update the prose and all relevant screenshots. We also add a couple of new screenshots and modify the prose to fix some inaccuracies and better showcase the installation of the first sample application using kubectl.


docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -7,54 +7,34 @@ Once you [have launched a new cluster](create-shoot-cluster.md), you can interac
...
 ## Extracting the kubeconfig file
 
-With the advent of Kubernetes 1.27, you can access your shoot cluster using a *certificate-based* kubeconfig file.
-You may also be familiar with *static* kubeconfig files.
-
-* A **static** kubeconfig contains a [static token](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#static-token-file), provides `cluster-admin` privileges to the corresponding shoot cluster, and never expires.
-In {{brand}}, for existing --- not new --- clusters with Kubernetes 1.26, you can access your shoot by generating a static kubeconfig.
-
-* A **certificate-based** kubeconfig is valid for a predefined amount of time, during which it can be used for accessing the corresponding shoot cluster with `cluster-admin` privileges.
+With the advent of Kubernetes 1.27, you can access your shoot cluster using a **certificate-based** kubeconfig file.
```
fghaas, line 16: I don't see the point of the "with the advent" prefix anymore. Thus, I'd put:

```suggestion
You can access your shoot cluster using a **certificate-based** kubeconfig file.
```

docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -66,9 +46,9 @@ Create a directory named `.kube` in your local user's home, then move the YAML f
 Rename the YAML to `config`, ending up with `~/.kube/config`.
 By default, `kubectl` searches for its configuration in `~/.kube/config`.
 If necessary, you can [modify this behavior](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/) by setting the `KUBECONFIG` environment variable.
-For example, if you want to retain the original filename exactly as downloaded, you will type something like the following:
+For example, if you want to retain the original filename exactly as downloaded, you might type something like the following:
 
-```shell
+```console
```
fghaas, line 51: `shell` is for highlighting shell commands. `console` is for console screen dumps.

So, either leave `shell`, or else use `console` and add a console prefix like `$`.

I'd prefer to keep `shell`.

docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -78,65 +58,43 @@ If you manage multiple Kubernetes clusters, you will probably prefer this approa
 
 Check if your `kubectl` uses the proper configuration by running:
 
-```shell
+```console
```
fghaas, line 61: Here, too.

docs/howto/kubernetes/gardener/kubectl.md
```diff
@@ -146,122 +104,97 @@ They *also* disappear when you [hibernate the shoot cluster](hibernate-shoot-clu
 
 Check your available nodes by running:
 
-```shell
+```console
```
fghaas, line 107: ... and here.

fghaas (commented):
Looks good in general; I have a few notes.

docs/howto/kubernetes/gardener/kubectl.md
colder-is-better: I am switching to `shell`. By thw way, first I tried `console` with the `$` prefix, but `tox -e markdownlint` doesn't like the fact there's no command output, so it aborts like so: 

```
MD014: Dollar signs used before commands without showing output (commands-show-output)
```

fghaas (approved):
Looks good to me now. Any second thoughts, or is this OK to be merged?

colder-is-better:
As it is, I cannot think of a reason not to merge. Thanks!

PR #300, opened 2024-10-03
fix: Refresh rolling-upgrade shoot How-To guide

colder-is-better:
We support new Kubernetes versions, and we dropped support for static kubeconfig files. Thus, we refresh all screenshots and modify the prose accordingly.


docs/howto/kubernetes/gardener/rolling-upgrades.md
```diff
@@ -4,25 +4,28 @@ description: How to manually upgrade machine images or Kubernetes versions for G
 # Conducting rolling upgrades
 
 By default, Kubernetes clusters created with {{k8s_management_service}} in {{brand}} are upgraded automatically.
-Those upgrades take place during a specified maintenance window, and you may find out more about [what they involve and how they work](../../../background/kubernetes/gardener/autoupgrades.md).
+Those upgrades happen during a specified maintenance window, and you may learn more about [what they involve and how they work](../../../background/kubernetes/gardener/autoupgrades.md).
```
fghaas, line 7: I think "take place" is better, because it comes with a connotation of planning and premeditation. "Happen" sounds like it's random and arbitrary.

fghaas (commented):
Looks good — I just have one micro-nit. :)

docs/howto/kubernetes/gardener/rolling-upgrades.md
colder-is-better: Huh, I thought I would take the phrasal verb out but failed to realize what you're pointing out.

fghaas (approved):

PR #301, opened 2024-10-04
fix: Update shoot creation How-To guide for new parameters

colder-is-better:
Two new parameters are visible while creating a new shoot cluster. Those are related to the load balancer provider and the availability zones. At the moment, they are both preset and cannot be changed. Because of that, we update two screenshots and modify the prose accordingly.


fghaas (approved):

fghaas:
I've rebased the topic branch on current `main`, and I'm about to merge this.

colder-is-better:
Ack, thanks.

PR #309, opened 2024-12-06
fix: Update the limitations page to include volume backups

colder-is-better:
The volume backup service is not configured in any of the Cleura Cloud regions, so we update the limitations page accordingly.

While at it, we suggest configuring servers for Disaster Recovery via the Cleura Cloud Management Panel.

### DIFF
--- a/docs/reference/limitations/openstack.md
+++ b/docs/reference/limitations/openstack.md
@@ -20,6 +20,14 @@ It also means that you cannot use the following `openstack` CLI commands; they a
   > You can use `openstack project list`, but this will only list the project(s) that your user account has access to.
 * `openstack domain <subcommand>`
 
+## Cinder
+
+### Volume backup service
+
+{{brand}} does not support the OpenStack volume backup service (`cinder-backup`).
+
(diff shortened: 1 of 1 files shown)


docs/reference/limitations/openstack.md
```diff
@@ -20,6 +20,15 @@ It also means that you cannot use the following `openstack` CLI commands; they a
   > You can use `openstack project list`, but this will only list the project(s) that your user account has access to.
 * `openstack domain <subcommand>`
 
+## Cinder
+
+### Volume backup service
+
+The volume backup service is not configured in any of the {{brand}} regions.
+As a result, volume backups cannot be created.
```
fghaas, line 28:
```suggestion
{{brand}} does not support the OpenStack volume backup service (`cinder-backup`).

For automated, scheduled volume *snapshots,* consider configuring your servers for [Disaster Recovery (DR)](../../background/disaster-recovery.md) via the {{gui}}.
```

fghaas (commented):
I have proposed an alternate wording for this. Let me know what you think.

docs/reference/limitations/openstack.md
colder-is-better: It is ever so slightly better than my wording, so let's go ahead with yours :)

fghaas (approved):

PR #313, opened 2025-03-04
fix: List object storage in Sto2 as "in deployment"

fghaas:
Object storage support in Sto2 is coming, though it's not ready for general consumption yet.


fghaas:
@colder-is-better FYI, I do not know why the build is currently failing, but it appears to be a transient error. I'll try re-running the checks later.

colder-is-better (approved):

PR #316, opened 2025-04-02
fix: Replace Legal page content with link to authoritative source

fghaas:
All legally relevant documents are now at https://cleura.com/resources/legal/terms/. Thus, for avoidance of ambiguity, refer to just that page in the Legal Reference subsection.

Also, remove the extra variables containing direct links to the PDFs, and consequently also remove references to those variables from the account creation how-to guide.


colder-is-better (approved):
LGTM, thanks!

PR #321, opened 2025-04-04
fix: Update OpenStack version information for Sto-Com

fghaas:
The Sto-Com region is running OpenStack Caracal, not Antelope.
Update the version reference accordingly.


colder-is-better (approved):

noonedeadpunk (approved):
LGTM

PR #329, opened 2025-04-25
Feat/admonition

AboveHero:
This PR enables the use of admonitions, which offer a modern and contrasted way to highlight information.

### 1.
Screenshot from 2025-04-25 14-57-40

### 2.
Screenshot from 2025-04-25 14-57-58


fghaas:
We used to have admonitions, and removed them in d62071a54787b1cdd29853e2d89a87c8b1de0bf2 for the reasons outlined in that commit message. You'll have to convince me that those concerns no longer apply. :)

PR #330, opened 2025-05-05
Update volume type reference

albgus:
Updated the volume type reference to more accurately represent the types exposed in the API. Specifically made sure that the separate volume types with (`-encrypted`) is represented in this table, and that the default type is not encrypted. I am assuming that since there are volume types specifically ending in `-encrypted` the ones without the suffix is unencrypted.

### DIFF
--- a/docs/reference/volumes/index.md
+++ b/docs/reference/volumes/index.md
@@ -7,10 +7,12 @@ The following volume types are available in {{brand}} for persistent block stora
 If you create a volume without specifying a volume type, then the default volume type applies.
 
 
-| Volume type    | Default          | [Encryption](../../howto/openstack/cinder/encrypted-volumes.md) |
-| -----------    | -------          | ------------                                                    |
-| `cbs-standard` | :material-close: | :material-check:                                                |
-| `cbs-premium`  | :material-check: | :material-check:                                                |
+| Volume type              | Default          | [Encryption](../../howto/openstack/cinder/encrypted-volumes.md) |
+| -----------              | -------          | ------------                                                    |
+| `cbs-standard`           | :material-close: | :material-close:                                                |
+| `cbs-premium`            | :material-check: | :material-close:                                                |
+| `cbs-standard-encrypted` | :material-close: | :material-check:                                                |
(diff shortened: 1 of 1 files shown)


docs/reference/volumes/index.md
```diff
@@ -7,10 +7,12 @@ The following volume types are available in {{brand}} for persistent block stora
...
 
 
-| Volume type    | Default          | [Encryption](../../howto/openstack/cinder/encrypted-volumes.md) |
-| -----------    | -------          | ------------                                                    |
-| `cbs-standard` | :material-close: | :material-check:                                                |
-| `cbs-premium`  | :material-check: | :material-check:                                                |
+| Volume type              | Default          | [Encryption](../../howto/openstack/cinder/encrypted-volumes.md) |
+| -----------              | -------          | ------------                                                    |
+| `cbs-standard`           | :material-close: | :material-close:                                                |
+| `cbs-premium`            | :material-check: | :material-close:                                                |
```
fghaas, line 13: `cbs-premium` actually _isn't_ the default volume type as defined in the Cinder API (that's `cbs-standard`), but it is indeed being selected by default in Cleura Cloud Management Panel. That happens to be a UI bug in CCMP, thanks for pointing that out! We'll make sure that it gets fixed.

fghaas (commented):
@albgus Thank you for the PR! If you could, please fix your commit message so that it passes the `gitlint` checks. This includes adding the appropriate prefix (in this case, `fix:`) to the subject line, formatting your commit message for an 80-character line length, and using the imperative mood. Like so:

```
fix: Update volume type reference
    
Update the volume type reference to more accurately represent the
types exposed in the API. Specifically, make sure that the separate
volume types with ("-encrypted") are represented in this table, and
that the default type is not encrypted.
```

If you prefer that we take of that instead, just let me know. In that case, we'd amend your commit accordingly.

fghaas:
Hello @albgus, we didn't hear back from you so we've applied the changes the changes I had requested, added a `Co-authored-by` credit, and merged the change in #331. Thank you!

PR #338, opened 2025-06-04
fix: Update security groups guide for IPv6

colder-is-better:
- Add more information regarding IPv6 connections
- Introduce additions & modifications for more detailed explanations
- Augment command examples & outputs to reflect IPv6-related operations
- Replace old screenshots & add new ones for better showcasing CCMP operations


colder-is-better:
Maybe that's a feature? Some PNGs are large enough or something, so they are not rendered by default?

PR #343, opened 2025-06-10
Add Marketplace How-To guides and background explainer

colder-is-better:
For each application in the Marketplace, we have a deployment along with a deletion guide.


fghaas (approved):
I've taken the liberty to change the PR title since it's not just how-to guides, but a background explainer as well.

I think this is good to go as-is, however please see the inline note re screenshots.

PR #346, opened 2025-06-24
Update limitations, add new guide regarding custom images

colder-is-better:
- Update the limitations document regarding the Glance web-download method,
- add a guide regarding custom image uploads via the OpenStack CLI, and
- show how to create custom images from server root volumes.


fghaas (approved):
Ths looks good to me, thanks!

PR #349, opened 2025-07-02
chore: One sentence per line for How-to guides index 

fghaas:
Reformat the "About our How-To guides" page to one sentence per line.


colder-is-better (approved):
Emacs be crazy.

PR #354, opened 2025-07-03
chore: Apply the one sentence per line policy in the s3 subcategory

colder-is-better:
We reformat all How-Tos and index files in the s3 subcategory of the object-storage category, to follow the one sentence per line policy.

### DIFF
--- a/docs/howto/object-storage/s3/object-lock.md
+++ b/docs/howto/object-storage/s3/object-lock.md
@@ -4,10 +4,8 @@ description: Object lock enables you to keep S3 objects from being deleted or ov
 # Object lock
 
 
-Object lock is a feature in S3 that enables you to lock your objects
-to prevent them from being deleted or overwritten. This feature can be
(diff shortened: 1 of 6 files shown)


docs/howto/object-storage/s3/object-lock.md
fghaas, line 1: I can't add this comment to the correct line because GitHub is silly that way, but there's still a wrapped sentence starting with "To configure your bucket to use the **Compliance** mode" in this file.

fghaas (commented):
Looks good; I have one tiny nit.

(The build failure was apparently due to an intermittent glitch with <https://docs.openstack.org/swift/>, and should hopefully resolve itself on the next build.)

docs/howto/object-storage/s3/object-lock.md
colder-is-better: Thanks -- I took care of that.

fghaas (approved):

PR #355, opened 2025-07-03
chore: Apply the one sentence per line policy in the swift subcategory

colder-is-better:
We reformat all How-Tos and index files in the swift subcategory of the object-storage category, to follow the one sentence per line policy.


fghaas:
Merged to `main` and from there (resolving a trivial merge conflict) to `v2`.

PR #359, opened 2025-07-04
chore: Apply the one sentence per line policy in the nova subcategory

colder-is-better:
We reformat all How-Tos in the nova subcategory of the openstack category, to follow the one sentence per line policy.


fghaas (approved):

colder-is-better:
@fghaas Ack, thanks!

PR #365, opened 2025-07-28
Merge main into v2

colder-is-better:


fghaas:
I am not at all sure where the "This branch cannot be rebased due to conflicts" message came from earlier, because no such conflict existed. At any rate, merging this worked just fine.

PR #368, opened 2025-08-08
fix: Refresh command outputs

colder-is-better:
We update console outputs of the openstack and swift clients.


colder-is-better:
Again, it's kubernetes-dot-io.

PR #370, opened 2025-08-12
Introduce per-region docs

fghaas:
This is a rather large and involved change that serves the following purpose: henceforth, rather than having all documentation for Cleura Cloud with substantially identical content at [docs.cleura.cloud](https://docs.cleura.cloud), we will have individual docs for each Cleura Cloud region at [docs.cleura.cloud/fra1](https://docs.cleura.cloud/fra1/) etc.

For this, we'll use the Mike version provider for MkDocs-Material.

Initially, the non-versioned content remains in the gh-pages branch. Only at a later stage will we revert [420cbd3](https://github.com/citynetwork/docs/pull/370/commits/420cbd386e6400c9005a5e91d2c406480316fdbf), and thus create a redirect from the old content to the new.

### DIFF
--- a/scripts/mike.sh
+++ b/scripts/mike.sh
@@ -17,4 +17,5 @@ mike deploy \
      $mike_version_alias \
      "$@"
 
-mike set-default $mike_default_version_name "$@"
+# TODO: Re-enable when deploying the mike-versioned docs to production
+# mike set-default $mike_default_version_name "$@"


colder-is-better (approved):
I don't see anything off, so it *should* be good to go.

fghaas:
> I don't see anything off, so it _should_ be good to go.

OK thanks; still, I'd like for us to merge #372 first, whereupon I will rebase this on then-current `main` one more time.

colder-is-better (approved):

PR #375, opened 2025-08-20
refactor: Invoke Mike using a Python, not a shell script

fghaas:
This enables us to subsequently get cleverer with setting the version
names and aliases.


colder-is-better (approved):
LGTM, thanks.

PR #376, opened 2025-08-20
fix: Replace hardcoded references to Sto-Com

fghaas:
Replace a few remaining references to `Sto-Com` with an `api_region` variable reference.


colder-is-better (approved):
LGTM -- thanks!

PR #391, opened 2025-08-29
fix: Update documentation in light of OpenStack Epoxy

colder-is-better:
Fra1 is the first region we run Epoxy, so we update the documentation accordingly.


fghaas (approved):
Looks good to me, thank you!

PR #393, opened 2025-09-02
Add guide on changing the password of an OpenStack API user

colder-is-better:
We show how to change the API user password via the Cleura Cloud Management Panel, or with the help of the OpenStack API client.


fghaas (approved):
Looks good to me, thank you!

PR #395, opened 2025-09-08
feat: Add missing index pages in howto section

fghaas:
Without this change, multiple subsections in our howto section lack an index page. This is ugly and not very user-friendly, so add them.


colder-is-better (approved):
LGTM -- thanks!

PR #396, opened 2025-09-08
fix: Add overview information for Sto-Com to all regions

fghaas:
With the per-region format, it no longer makes sense to restrict reference information about features and versions in Sto-Com to just that region.

Thus, add that information to the main branch.


colder-is-better (approved):
LGTM, thanks!

PR #405, opened 2025-10-01
fix: Correct gui_domain and rest_api_domain for Compliant Cloud

fghaas:
With per-region documentation, it makes more sense to directly use the CCMP and CCMS domain names for Compliant Cloud, in the relevant regions.


colder-is-better (approved):

fghaas:
@colder-is-better I had to make one other change to un-break the build checks.

colder-is-better (approved):

PR #428, opened 2025-10-24
feat: Expand DNS feature reference in Compliant Cloud

fghaas:
Designate is now in beta in all Compliant Cloud regions, including Sto-Com.


colder-is-better:
@fghaas Ack, thanks.

PR #436, opened 2025-11-25
feat: Add new Designate reference section

colder-is-better:
Each Designate zone comes with two name servers. The names of those servers seem to indicate the regions they are in. This is not the case, so we provide a matrix showing which region each name server resides in.

### DIFF
--- a/docs/howto/openstack/designate/recordsets.md
+++ b/docs/howto/openstack/designate/recordsets.md
@@ -158,6 +158,9 @@ nuuk.example.com.  3600    IN  AAAA    2001:db8:ffff:ffff:ffff:ffff:ffff:ffff
 ;; MSG SIZE  rcvd: 78
 ```
 
+> For extra resilience, the two name servers in any zone reside in different geographic locations.
+> You may see the location of a name server [in the Reference section](../../../reference/designate/index.md).
+
 ## Deleting recordsets
 
 To delete a recordset, you use the `openstack recordset delete` command and specify the name of the zone, together with the `id` of the recordset.
--- a/docs/reference/designate/index.md
+++ b/docs/reference/designate/index.md
@@ -1,0 +1,23 @@
+# DNS servers
+
+## Server regions
+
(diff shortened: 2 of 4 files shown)


docs/reference/designate/index.md
```diff
@@ -0,0 +1,22 @@
+# DNS servers
+
+## Server regions
+
+By default, each new zone you create with Designate comes equipped with two name servers.
+The names of those servers, like `cloud-ns1.fra1-pub.cleura.cloud.` and `cloud-ns2.kna1-pub.cleura.cloud.`, seem indicative of the {{brand}} regions they reside in.
+This is not the case, and the following matrix shows the regions each name server resides in.
```
fghaas:
```suggestion
The names of those servers, like `cloud-ns1.fra1-pub.cleura.cloud.` and `cloud-ns2.kna1-pub.cleura.cloud.`, are indicative of the regions whose DNS zones and records they handle.

The following matrix shows where each name server is located.
```

fghaas (commented):
Couple of notes. Also, the how-to section still needs a cross-reference to this new part of the reference.

colder-is-better:
I took into account the notes and also added a cross-reference in the How-To section.

docs/howto/openstack/designate/recordsets.md
```diff
@@ -158,6 +158,9 @@ nuuk.example.com.  3600    IN  AAAA    2001:db8:ffff:ffff:ffff:ffff:ffff:ffff
 ;; MSG SIZE  rcvd: 78
 ```
 
+> For extra resilience, the two name servers in any zone reside in different geographic locations.
+> You may see the location of a name server [here](../../../../reference/designate/).
```
fghaas: Fix the build failure, please.

Also, improve accessibility by using a link text other than "here". Like so, for example.

```suggestion
> You may see the location of a name server [in the Reference section](../../../../reference/designate/index.md).
```

fghaas (approved):

PR #470, opened 2026-04-08
UI refresh

colder-is-better:


colder-is-better:
@fghaas This is a core dump, please ignore.

PR #488, opened 2026-04-24
feat: Add the invite-only admonition to all AI-related pages

fghaas:
Since the admonition at the top of the page functions somewhat akin to a banner, add it to **all** AI-related pages, as opposed to just the index pages of the affected sections.


colder-is-better (approved):
LGTM, thanks!

PR #490, opened 2026-04-27
Recreate screenshots in the Marketplace subsection, in light of the new CCMP theme

colder-is-better:
Re-create all screenshots to ensure proper switching between light and dark themes.
Also, avoid mentioning UI colors in the prose.


fghaas (approved):
Great work. Thanks!

PR #491, opened 2026-04-28
fix: Remove references to Clavister NetWall

fghaas:
Drop the references to Clavister Netwall until infrastructure issues in Sto-Com have been resolved.


colder-is-better (approved):
Everything that had to be removed seems removed, so this LGTM.

PR #502, opened 2026-05-27
chore: Add further "public" placeholders

fghaas:
Set the placeholders/markers for the public branch to the theme reference and canonical version.
    
Reset api_domain to Kna1, so that the link check does not break on PRs.

(This PR primarily exists to test the protected branch workflow on this new branch.)


colder-is-better (dismissed):
Changes make sense -- thanks!

colder-is-better (approved):

PR #509, opened 2026-06-02
Add new How-To guide regarding keypair management

colder-is-better:
* Add a new step-by-step guide on keypair creation, viewing, duplication, and deletion.

* Remove instructions regarding keypairs from server creation guide.

* Standardise on the use of "keypair", rather than "key pair".

* Add a subsection heading in the "Accessing the OpenStack API" guide for sourcing the RC file, to facilitate linking.


fghaas (approved):
Looks good to me, thanks!

PR #518, opened 2026-06-07
Recreate screenshots in the Getting Started subsection, in light of the new CCMP theme

colder-is-better:
Re-create all screenshots to ensure proper switching between light and dark themes.

Also, avoid mentioning UI colors in the prose, and use Material icon codes to refer to UI elements.


fghaas (approved):
Rebased this on current `main`. Looks good to me, thanks!

PR #527, opened 2026-06-15
Update docs in light of the latest supported Kubernetes version

colder-is-better:
Presently, our Gardener clusters support Kubernetes releases up to 1.35.

Update the public documentation to reflect that.


fghaas (approved):
Looks good to me, thanks!

